PR #13, opened 2019-09-08
Typo in description function SUMIFS (Calc)

DiGro:
In the file helpcontent2/source/text/scalc/01.po  
location: 04060106.xhp#par_id3152148K.help.text

In line 1002 slipt in a single typo.  
The word "entires" (without qoutes)  should be "entries"  (without qoutes)


Pilot-Pirx:
> In the file helpcontent2/source/text/scalc/01.po
> location: 04060106.xhp#par_id3152148K.help.text
> 
> In line 1002 slipt in a single typo.
> The word "entires" (without qoutes) should be "entries" (without qoutes)

Hi Dick,

Thanks for the finding, I will correct it...

You sure don't want to merge  206 commits into trunk from AOO42X ;-)

Regards,
   Matthias

Pilot-Pirx:
Committed to trunk and cherry-picked for AOO42X.

Thanks for your contribution!

Pilot-Pirx:
Of course this change will be only visible in en-US at the moment.
Translation for other languages must be done in Pootle, after we have generated a new template. Then localization can be exported as an SDF file and committed to code.

DiGro:
Mathias,
thanks for the commit

I stumbled over this one when I was looking for something else in Pootle.
I saw that it was already often corrected in other languages :-)They probably missed it or didn't notice it.
So, no problem if it takes some time


__________
DiGro 

    Op zondag 8 september 2019 18:12:04 CEST schreef Matthias <notifications@github.com>:  
 
 
Of course this change will be only visible in en-US at the moment.
Translation for other languages must be done in Pootle, after we have generated a new template. Then localization can be exported as an SDF file and committed to code.

PR #15, opened 2019-10-07
Issue 97058 - Add Fontwork into Insert menu

oooforum:
Read: https://bz.apache.org/ooo/show_bug.cgi?id=97058 for more info


cbmarcum:
I have built the office with this patch and it adds the Insert > Object > Fontwork Gallery to Writer as described.

cbmarcum (approved):
I have built and tested this patch and it adds the menu item Inset > Object > Fontwork Gallery to Writer.

Pilot-Pirx (approved):
I don't see anything that speaks against adding this to the menu.

Maybe we should add it to other menus as well, but that can be discussed later.

PR #23, opened 2020-01-10
Update Python to version 2.7.17

pgiffuni:
This is an update to python given that AOO still ships with 2.7.15 and there is a huge number of issues with it. See for the release notes:

https://www.python.org/downloads/release/python-2717/

Note that I personally use the external version but the patches here were tested to apply.
I also reduced greatly the FreeBSD patch, since upstream is not supporting FreeBSD 9 anymore.


pgiffuni (commented):
Oops! Please ignore the unrelated changes in external_deps.lst, those were left-overs from my older trees.

Only python is updated.

Pilot-Pirx:
Thank you for your contribution!

PR #27, opened 2020-01-25
Partial fixes to support Python 3

pgiffuni:
I authorized Damjan@ to commit these already but gor the record.


pgiffuni:
These are some of the pyuno changes required to compile using Python 3.
In some cases this patches actually switches to the pre-python 3 code that was somewhat more portable.

DamjanJovanovic:
Merging. Thank you for your contribution!

PR #31, opened 2020-02-12
Fix for issue #127895

oooforum:
See https://bz.apache.org/ooo/show_bug.cgi?id=127895


leginee:
How do we test it if it is right?

cbmarcum:
This looks legitimate. I'd like someone better than me at Cpp to look at it.

PatriciaShanahan:
It is difficult to verify experimentally, because the symptom would be failure to dispose of an object that is no longer needed.

From a code inspection point of view, it looks valid to me. It appears to be a copy-paste-edit error, in which an edit was missed. The block for disposing of xListener was probably copied while writing the block for xDocListener, but one instance of xListener was not changed to xDocListener.
```
else if ( xListener.is() )
      m_pData->m_aInterfaceContainer.removeInterface( ::getCppuType((const uno::Reference< XEVENTLISTENER >*)0), xListener );
else if ( xDocListener.is() )
      m_pData->m_aInterfaceContainer.removeInterface( ::getCppuType((const uno::Reference< XDOCEVENTLISTENER >*)0), xListener );
```

cbmarcum (approved):
I agree this looks like a copy/paste error and the correct listener should get removed.

PR #32, opened 2020-02-12
Fix for issue #127897

oooforum:
See https://bz.apache.org/ooo/show_bug.cgi?id=127897


ardovm:
Please see my comment on the bug report.
Do we still believe this is a bug?

leginee:
I am not sure if this is a bug or not. We should to investigate this further.

ardovm:
@oooforum if you could follow the discussion on the BugZilla report, this "exchange" of auto-grow height and auto-grow size may be intentional.
If it was so, then this PR should better add clarifying comments, rather than change the code's behavior.
Do you have any opinions about this?

PR #37, opened 2020-02-22
Removed whitespace between tags bookmark_value

DiGro:
One instance removed which disrupted the flow of view and were probably placed for reasons of readability in the original code files.
These superfluous whitespace hindered proper translation.


DiGro:

commit 97a1911f333e8e743f083780d3b1eb0c0ea6a5ec

PR #73, opened 2020-02-22
Removed whitespace between tags bookmark_value

DiGro:
One instance removed which disrupted the flow of view and were probably placed for reasons of readability in the original code files.


DiGro:

commit 97a1911

PR #75, opened 2020-02-29
Removed whitespace from tags <bookmark_value> 20200229 1517

DiGro:
Removed instances from tags <bookmark_value> which disrupted the flow of view and were probably placed for reasons of readability in the original code files.


Pilot-Pirx:
Due to last comment:
Can this PR be closed then?

PR #76, opened 2020-02-29
Removed whitespace from tags <bookmark_value> 20200229 1835

DiGro:
Removed instances from tags <bookmark_value> which disrupted the flow of view and were probably placed for reasons of readability in the original code files.


leginee:
Is there a reason why multiple lines have now been addid into a single line?

DiGro:
Yes, but it is not that relevant.

When I first started to remove the whitespace, I noticed that the whitespace was
caused by extra tabs/spaces before the text. Text should be aligned to the left
margin to avoid superfluous whitespace.

In the beginning I removed the whitespace and put all text behind each other.
That solved the problem of the whitespace. Created however long lines, as you noticed ;- )

Later on I noticed I only had to remove the whitespace before the text so that
text was aligned at the left margin and all was OK. 

I used the last method for the remainder of the texts but yes, there are some long lines.
Since they are text that is included into an index for the help, it doesn't seem to
cause any problems anymore.
Whilst I was working on it, I noticed both long lines as well as short separate lines are used
in the original files.

__________
DiGro


     Op zondag 3 mei 2020 00:35:15 CEST schreef Peter <notifications@github.com>:  
 
 


Is there a reason why multiple lines have now been addid into a single line?

leginee:
I see. Thanks for the update.
So far I am not negative to this proposal, but I have not checked the change within OpenOffice.
Hope I can give feedback soon.

cbmarcum:
Where would I look in the office to see the effect of this change?

DiGro:
Carl,

most of these changes are not visible in the office suite itself.
At least I couldn't find them where I expected them

They are part of the index for the Help, but they show correct in the Help from the suite itself.

They caused problems however with translating when they are imported into Pootle.
Pootle is the program used for translations from en-us to other languages.

In Pootle they generate unwanted spaces in the strings. Because of that the strings
in Pootle generate in their turn warnings and failed checks.
These warnings and failed checks have to be corrected by the translators.

By removing these superfluous spaces strings will be correct in Pootle and won't raise warnings and failed checks anymore.

The superfluous spaces were, probably accidentally, inserted in the past when people, for reasons of readability,
added them to the original files that then were imported into Pootle and caused this problem, probably without the original
authors knowing of it.

Removing the white spaces will lighten the "burden" for translators to correct the caused problems.

Other removed white space were truncated strings (probably also for reasons of readability in the original file in the past) en
wrongful placed spaces (for instance at end of sentence but before end-tag).

One example of a truncated string in the Help for Calc is:
"Example: Cell E5 contains the text "Europe". Below, in cell E6, is the value 100 and in cell E7 the value 200. If the Automatically find column and row labels box is marked, you can write the following formula in cell A1: =SUM(Europe)."
which shows correct in the Help, but does not in Pootle.

In Pootle:
" <emph>Example</emph>: Cell E5 contains the text "Europe". Below, in cell E6, is the value 100 and in cell E7 the value 200. If the <emph>Automatically find column and row labels</emph> box is marked, you can write the following formula in cell A1: =SUM(Europe)."

Note the excessive spaces before <emph>Example</emph>

I suppose Mechtilde and Mathias have more insight on the technical details, since I'm not really a "techie"


__________
DiGro


     Op zondag 3 mei 2020 16:24:34 CEST schreef Carl Marcum <notifications@github.com>:  
 
 


Where would I look in the office to see the effect of this change?

cbmarcum (approved):
Thanks for the thorough explanation.
In the future I think the first bookmark_value could be pulled down onto it's own line for consistency. 
In this case I agree that pootle translation trumps technical readability of indenting the bookmark_value which is within the bookmark tag.

Pilot-Pirx:
I am in the massive process of cleaning up our Help files...
The whitespace (not only) before <bookmark_value> should be gone now.
Even if we don't use Pootle in the future a clean structure is needed.

When all files from this PR are updated I will close it.

BTW: To maintain Help (.xhp) files I packaged our Help Authoring tool:
https://home.apache.org/~mseidel/HelpAuthoring.oxt

However it surely needs some attention from a developer (Open Help file is still missing)

Pilot-Pirx:
Closing this PR as all files are updated now.

PR #74, opened 2020-02-28
Removed whitespace between tags <bookmark_value> and other superfluous whitespace (branch Latest)

DiGro:
Commits contain removed whitespace from tags <bookmark_value> and some other odd instances encountered while working on the files.
The commits with .gitattirbutes should be ignored though. One is for creating the file and one for deleting it.  I don't seem to be able to delete these twoo entries  .gitattributes. Sorry for that

PR #80, opened 2020-03-07
Fixed - single file patches  combined into one PR and commit 97a1911

DiGro:
After polluting this repo with a great number of Pull Request with just the one file attached, I managed to put all of them into just the one commit 97a1911 which is attached to this PR.

I will close the earlier made PR's  33 to 73 since those are clustered in this one

https://github.com/apache/openoffice/commit/97a1911f333e8e743f083780d3b1eb0c0ea6a5ec


dave2wave:
Does this patch replace #76, #75, and #74?

DiGro:
@dave2wave : "Does this patch replace #76, #75, and #74?"

No, it doesn't. Those are subsequent patches regarding the superfluous spaces in the Help files which show up in the translation tool Pootle. These patches fix that.

I was then (march 2020) just starting out with Github.  Didn't know if it could handle so much fixes in one patch, so I split them up. There were very much files to fix. Hence the patches #74, #75 and #76.

robertaq1998 (approved):

Pilot-Pirx:
Hi Dick,
thank you for all this work. I think I have updated all files from this PR with my big "Cleanup". So I will close it now.

However, there are still a lot of small errors in the Help files which we can solve later. Hopefully we will then have a working translation engine available.

DiGro:
@Pilot-Pirx 
Matthias, thanks for your tremendous efforts.
I'm not sure how to help you with the Help files, but if there is something I can do, please let me know.
Frohes neues Jahr :-)

Pilot-Pirx:
Happy New Year to you, too!

I think I will finish this week, after that we can go back to normal maintenance of the Help files. I will let you know.

PR #81, opened 2020-03-18
Update README.md

samuel100u:
Hi,

Change `source *.set.sh`  to `source *.Set.sh`

messageImage_1584511779997


Pilot-Pirx:
Thank you for finding this detail!

PR #86, opened 2020-04-28
Translation

leginee:
I translated some german comments to english in some files.
My Idea what would be a benefit to review:

- there are only changes in comments 
- spelling
- if the translations make sense (whether you check against code or against the german does not matter)


Pilot-Pirx:
I did some proofreading and I think this could be merged now...

Pilot-Pirx (approved):

cbmarcum (commented):
Hi Peter,
Here is what I found.
main/sc/source/ui/app/client.cxx

line 223
-- adopt sinze into the document
++ adopt size into the document


main/sfx2/source/appl/appopen.cxx 
line 151
-- Name of the Dockuments including path
++ Name of the Documents including path

line 152
-- sal_True: do not ask vor new view
++ sal_True: do not ask for new view

line 300
-- Medium that needs a passwort
++ Medium that needs a password

lines 304-307
--
To query a passwort on a mediom works only if the medium is a storage.
if in documentinfo the password-flag is set, a dialog will querry user 
for the poassword. The password will be saved in a set. If the set does not 
exist, a set will be created.
++
To query a password on a medium works only if the medium is a storage.
if in documentinfo the password-flag is set, a dialog will query user 
for the password. The password will be saved in a set. If the set does not 
exist, a set will be created.

line 578
-- use filename instaed of region/template
++ use filename instead of region/template

line 825
-- now check wether a stream is already there
++ now check whether a stream is already there

Pilot-Pirx:
Hi Carl,

I think I already corrected these typos, can you have a second look?

PR #87, opened 2020-06-07
Stax removal from source tree

leginee:
As suggested in the [Discussion on dev](https://lists.apache.org/thread.html/rec5be7b838a16c55ee64c12128dddd515bc10dc74c20b3f5bf57e85e%40%3Cdev.openoffice.apache.org%3E), I have created a patch that removes Stax.

also see [Bug 128383](https://lists.apache.org/thread.html/rec5be7b838a16c55ee64c12128dddd515bc10dc74c20b3f5bf57e85e%40%3Cdev.openoffice.apache.org%3E)

My testing has been so for:
- a successfull build on Linux
- executed a function in OpenOffice that is affected (see bug comment 3)


DamjanJovanovic:
There's a dependency from filter -> saxon -> stax

```
$ grep SAXON filter -R
filter/prj/build.lst:fl      filter  :    L10N:l10n svtools unotools xmloff cppu tools cppuhelper sal svx javaunohelper jvmaccess canvas SAXON:saxon LIBXSLT:libxslt basegfx NULL
filter/source/xsltfilter/makefile.mk:.IF "$(DISABLE_SAXON)" == ""
filter/source/xsltfilter/makefile.mk:.IF "$(SYSTEM_SAXON)" == "YES"
filter/source/xsltfilter/makefile.mk:XCLASSPATH:=$(XCLASSPATH)$(PATH_SEPERATOR)$(SAXON_JAR)
```

Apparently the xsltfilter uses it. Not sure where xsltfilter is used but maybe this helps:
https://wiki.openoffice.org/wiki/Documentation/DevGuide/OfficeDev/Integrating_Import_and_Export_Filters

leginee:
Yes that is correct. Let me Quote Pedro Giffuni from the discussion:

```
While looking for candidates to remove stuff, consider stax. It is
included in Java since JDK5.
```

So I tried if something breaks. XSLTFILTER is used only in the MS Docuxment import I think, so I loaded a document with my testbuild of AOO and did so far not find any issues.

leginee:
i will close the pull request. We have never integrated it. now it is obsolete. Since i migrate to bazel, and i marked it as not to migrate.

PR #92, opened 2020-08-21
AOO418 openssl upgrade

DonLewisFreeBSD:
Upgrade to openssl-1.0.2t.


Pilot-Pirx:
Shouldn't that be 1.0.2t in /main/openssl/version.mk?
(The same applies to trunk and AOO42X)

PR #93, opened 2020-08-23
Force Linux builds to use gnu++98 (C++98 + GNU extensions) mode to be

DonLewisFreeBSD:
Force Linux builds to use gnu++98 (C++98 + GNU extensions) mode to be
the most compatible with our old code base rather than relying on the
compiler default mode.  Compiling in C++11 or newer mode is very noisy
due to deprecation warnings about our use of std::auto_ptr.  If the
compiler defaults to C++17 mode or newer, the build would be totally
broken because std::auto_ptr is removed from C++17.

There is an unknown amount of porting effort needed to convert to
std::unique_ptr, which has somewhat different semantics and which is
not available before C++11, which would break building with older versions
of gcc which default to gnu++98 mode.

std::shared_ptr, might be an alternative since there is a BOOST
implementation.

This reduction of compiler warnings reduces the size of the build log by
2/3rd when compiling with gcc 6.3.0.

### DIFF
--- a/main/solenv/gbuild/platform/linux.mk
+++ b/main/solenv/gbuild/platform/linux.mk
@@ -89,6 +89,7 @@ gb_CXXFLAGS := \
 	-fuse-cxa-atexit \
 	-fvisibility-inlines-hidden \
 	-fvisibility=hidden \
+	-std=gnu++98 \
 	-pipe \
 
 ifneq ($(EXTERNAL_WARNINGS_NOT_ERRORS),TRUE)
--- a/main/solenv/inc/unxlng.mk
+++ b/main/solenv/inc/unxlng.mk
@@ -88,7 +88,7 @@ CFLAGSEXCEPTIONS=-fexceptions -fno-enforce-eh-specs
 CFLAGS_NO_EXCEPTIONS=-fno-exceptions
 
 # -fpermissive should be removed as soon as possible
-CFLAGSCXX= -pipe $(ARCH_FLAGS)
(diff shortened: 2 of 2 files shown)


leginee:
How about we define a stlport::unique_ptr?
We use
auto_ptr for older builds and unique_ptr for newer.

but basically I am fine in general with building with gnu++98

DonLewisFreeBSD:
> How about we define a stlport::unique_ptr?
> We use
> auto_ptr for older builds and unique_ptr for newer.

They are similar but unique_ptr is not a drop in replacement, otherwise they would have kept the same name.  I'm assuming that there would be some amount of porting and testing effort involved in switching.

I don't thing gcc is new enough to support unique_ptr for anything older than CentOS 7.  I'm pretty sure that we would have to switch to a newer version of Visual C++ as well.

leginee:
We need to find a method to do the modernization of the code. If we define this in stlport then we would have one place to mess around with. Also the idea is we could keep the support for older compilers while we modernize the code base as long as possible.
According to [the c++ Reference for auto_ptr](https://www.cplusplus.com/reference/memory/auto_ptr/) the unique_ptr is the replacement:

>Note: This class template is deprecated as of C++11. unique_ptr is a new facility with a similar functionality, but with improved security (no fake copy assignments), added features (deleters) and support for arrays. See unique_ptr for additional information.

If we use fake copy assignments we would need to replace them, while arrays we would not use the unique_ptr anyhow.

PR #95, opened 2020-09-17
JDK8 ODK

DonLewisFreeBSD:
Correct javadoc comments to unbreak --enable-odk build with JDK 8 or newer

### DIFF
--- a/main/javaunohelper/com/sun/star/comp/helper/Bootstrap.java
+++ b/main/javaunohelper/com/sun/star/comp/helper/Bootstrap.java
@@ -97,6 +97,7 @@ public class Bootstrap {
 		@param context_entries the hash table contains mappings of entry names (type string) to
 		context entries (type class ComponentContextEntry).
 		@return a new context.
+		@throws java.lang.Exception
     */
 	static public XComponentContext createInitialComponentContext( Hashtable context_entries )
         throws Exception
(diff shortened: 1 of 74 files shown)


Pilot-Pirx (commented):
My Windows test builds (incl. SDK) were successful.
But when building with Java 8 (Oracle) an installed Java 7 is not detected by AOO.
Since we still are building our releases with Java 7 this would be a minor regression for our users.

cbmarcum (approved):
built PR with JDK 8 on CentOS 7 and everything worked as expected.  Ran automated BVT tests and all 46 were passed.

DonLewisFreeBSD:
> My Windows test builds (incl. SDK) were successful.
> But when building with Java 8 (Oracle) an installed Java 7 is not detected by AOO.
> Since we still are building our releases with Java 7 this would be a minor regression for our users.

I believe the problem is present without the patch, but the build with Java 8 would have to be done w/o the SDK.

I will try to dig into this.

Pilot-Pirx:
> I believe the problem is present without the patch, but the build with Java 8 would have to be done w/o the SDK.
> 
> I will try to dig into this.

Yes, I think this is independent of the patch. A general problem we have with 4.1.x
Building AOO 4.2.x with Java 8 works as expected.

DonLewisFreeBSD:
> > I believe the problem is present without the patch, but the build with Java 8 would have to be done w/o the SDK.
> > I will try to dig into this.
> 
> Yes, I think this is independent of the patch. A general problem we have with 4.1.x
> Building AOO 4.2.x with Java 8 works as expected.

Which is odd.  A while back I compared the detection code between the two branches and they appeared to be identical.

cbmarcum:
I went back and checked my build and I can confirm what Matthias found about Java 7 was not discovered when I built with Java 8. But I don't think that is an issue for this PR.

PR #98, opened 2020-09-18
updated junit url to https for BVT test ant build

cbmarcum:
Ant build can't fetch junit since it requires https now.

501 HTTPS Required.
Use https://repo1.maven.org/maven2/
More information at https://links.sonatype.com/central/501-https-required


cbmarcum:
we may want to pull this into AOO418 also since you can't run BVT tests without it.

Pilot-Pirx:
I think this should be applied to all branches.

Maybe later we can move to junit-4.13.jar?

PR #100, opened 2020-09-29
Nssupgrade - Upgrade bundled nss to nss-3.39-with-nspr-4.20

DonLewisFreeBSD:
Upgrade the bundled version of nss to nss-3.39-with-nspr-4.20.  This is not the latest version, but newer versions use a lot of C90, which the old version of Visual C++ that we use can't handle.

nss contains security libraries and is used for document signing.


DonLewisFreeBSD:
This needs to be tested on MacOS.  Nss integrates somewhat strangely on that platform and we had issues with the previous upgrade.

Pilot-Pirx:
Just for the record:
Build on Windows was successful. This was Windows 10 64-bit with MozillaBuild 3.3 (64-bit)

cbmarcum:
I built this PR with trunk on CentOS 7. BVT tests and some manually file creation and testing went fine.
I don't think I have a certificate I can test signing a document with.
Is there anything specific I can do to test this more?

DonLewisFreeBSD:
I did my testing with a self-signed certificate that I created about a year ago.  The details are hazy now, but I found this set of instructions:
  https://www.dalesandro.net/create-self-signed-smime-certificates/

I then imported the .pkcs12 file into Firefox for signing documents.  OpenOffice uses the same certificate storage as Firefox.  I haven't tried it, but I would think that Chrome would do the same.

This isn't a great test because the certificate is self-signed, so OpenOffice can't validate it like it could if the certificate was issued by a well known certificate authority.  I'm hoping that someone who has an "official" certificate can do some testing.

DonLewisFreeBSD:
Another test would be to open an existing validly signed document and verifying that the signature is valid.

JohnRDOrazio:
I was reading in the recent threads that there is a MacOS machine with Catalina available for testing builds, perhaps this could be tested on that machine? I'm needing an updated libnss3.so and libnssutil3.so in order to be able to integrate the Chrome Embedded Framework into my Java add-on for Writer... CEF (or JCEF in the Java flavoured build) on Linux is requiring at least v3.22 . So updating to 3.39 would work great!

leginee:
> I was reading in the recent threads that there is a MacOS machine with Catalina available for testing builds, perhaps this could be tested on that machine?

yes, that is one purpose of the machine is to enable devs without a mac to test their work on the mac.

JohnRDOrazio:
I was just [reading up in bugzilla](https://bz.apache.org/ooo/show_bug.cgi?id=126891) on all the work done on testing this, thanks to don Lewis and to Kay for your effort in making this work!

Pilot-Pirx:
So basically, we are waiting for someone to do a build for macOS?

Pilot-Pirx:
If a signed document is needed to verify, feel free to take this one:
https://home.apache.org/~mseidel/Signiert.odt

DonLewisFreeBSD:
> So basically, we are waiting for someone to do a build for macOS?

Pretty much, although  I think someone reported problems with signature verification on MacOS even prior to this upgrade.

Pilot-Pirx:
I have now built again for Windows 10 (incl. the latest change) and it was successful.

Furthermore, my signed document can be verified with that build:
https://home.apache.org/~mseidel/Signiert.odt

Additionally I have manually manipulated above document, so this one should give a warning:
https://home.apache.org/~mseidel/Signiert%20-%20Korrupt.odt

Pilot-Pirx:
A basic test on macOS Catalina with AOO 4.1.10 could not validate the document signed with my (valid) certificate.
It looks like the root certificate is missing in the Certification Path on macOS.

ardovm:
With some edits, I could get these commits integrated into the "catalina" branch and built them successfully under macOS.
I had to make the following changes: http://home.apache.org/~ardovm/openoffice/catalina/catalina_nss.patch
(I am not sure what is the best way to add a patch here?)

A binary in dmg format is available [here](http://home.apache.org/~ardovm/openoffice/catalina/2021-09-08/).

Certification paths seem to remain not verifiable, unfortunately, as reported by @Pilot-Pirx.

I suggest @DonLewisFreeBSD integrate the patch into this branch and we check it did not break Linux and Windows builds, if you agree?

dave2wave:
In AOO41X I had to set an environment variable and open OpenOffice in a terminal window.

```
% MOZILLA_CERTIFICATE_FOLDER='sql:/Users/davewave/Library/Application Support/Firefox/Profiles/hk7tw91t.default-release-1631294076295'
% open -a 'OpenOffice'
```

I'm on macOS Catalina

ardovm:
We need this update. I am going to merge it now. Thank you, @DonLewisFreeBSD !

PR #102, opened 2020-09-30
Libxml+serf 418

DonLewisFreeBSD:
Bug fixes from upstream for bundled libxml2 and serf modules

* libxml2
  - Possible infinite loop in xmlStringLenDecodeEntities
  - Make sure that truncated UTF-8 sequences don't cause an out-of-bounds array access.
  - Fix memory leak in xmlSchemaValidateStream

* serf
  - Fix handling of NUL characters in certificate fields

### DIFF
--- a/ext_libraries/serf/NULbytes.patch
+++ b/ext_libraries/serf/NULbytes.patch
@@ -1,0 +1,313 @@
+------------------------------------------------------------------------
+r1699925 | breser | 2014-08-04 11:04:00 -0700 (Mon, 04 Aug 2014) | 5 lines
+
+On the 1.3.x branch: 
+
+Merge changes from trunk:
+- r2392: Handle NUL bytes in fields of X.509 certs.
+
+------------------------------------------------------------------------
+Index: misc/build/serf-1.2.1/buckets/ssl_buckets.c
+===================================================================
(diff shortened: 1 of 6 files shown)


Pilot-Pirx (approved):
Builds and installs fine on Windows. Did not show problems on limited tests.
However I wouldn't know how to test against the fixed problems.

cbmarcum (approved):
PR built on CentOS 7 VM.  I ran the automated build verification tests and all 49 passed.  did a few manual file create and edit type tests and it seems okay to me.

DonLewisFreeBSD:
Testing the serf bug fix would require making an SSL connection through a MITM device that redirected SSL network connections to intended to go to the server "example.com" to a rogue server that has a certificate for "example.com\0.badguy.com".  Without the fix, the connection would be allowed.  With the fix, the connection attempt should fail with a certificate error.

I don't have reproducers for the libxml2 fixes, but they would need to be embedded in a document and two of the bugs would cause a potential DoS (memory leak or infinite loop).

Since the patches came from upstream, I'm inclined to trust them as long as we don't see any regressions.  The libxml2 patches will be included in the next release.  The serf patch has been part of a released version of serf for many years.  Unfortunately upgrading to a fixed release of serf is non-trivial.

PR #103, opened 2020-10-03
Update ParseContext.cxx

Pilot-Pirx:
I think this is only one of those typical typos...
Using OpenGrok could not find any other reference outside this file, but I am not sure if this would have side effects.


DonLewisFreeBSD:
Looks fine to me.

PR #105, opened 2020-10-12
Fix build when configured with --without-fonts

DonLewisFreeBSD:
Packaging breaks when the --without-fonts option is passed to
configure because the fc_local.conf file is missing.

This seems to have been caused by a mismerge when cherry picking
commits from AOO42X.


cbmarcum (approved):
Successfully built this PR on CentOS 7 using --with-package-format="rpm deb" and --without-fonts and it packaged okay.

Pilot-Pirx:
Just FTR:
I built AOO for Windows with the patch applied and with release configuration.
No problems!

PR #106, opened 2020-10-16
Fix for crashing in exporting PDF with a Noto Serif CJK JP font.

androidports:
I got a bug report of crashing in exporting PDF with a Noto Serif CJK JP font from my Android user.
This is same bug in PC version:
I confirm this crash with:

OpenOffice 4.1.7
macOS 10.15.6

I fix this problem and send feed back patch.

The attached file is the sample document for crashing with exporting PDF.

[crashing-with-noto-serif-cjk-jp.docx](https://github.com/apache/openoffice/files/5390104/crashing-with-noto-serif-cjk-jp.docx)

### DIFF
--- a/main/vcl/source/fontsubset/cff.cxx
+++ b/main/vcl/source/fontsubset/cff.cxx
@@ -40,7 +40,7 @@ typedef long long S64;
 
 typedef sal_Int32 GlyphWidth;
 
-typedef float RealType;
+typedef double RealType;
 typedef RealType ValType;
 #include <vector>
 typedef std::vector<ValType> ValVector;


Pilot-Pirx:
Hi,
Can you please test with AOO 4.1.8-RC2?

https://dist.apache.org/repos/dist/dev/openoffice/4.1.8-RC2/binaries/

androidports:
OK, I confirm crashing is fixed in AOO 4.1.8-RC2.
But exported PDF is invalid. (attached file)

[crashing-with-noto-serif-cjk-jp.pdf](https://github.com/apache/openoffice/files/5390829/crashing-with-noto-serif-cjk-jp.pdf)

I think 4.1.8-RC2 avoid crashing by invalid CFF format, but this invalid format is caused by casting from `float` to `int` in main/vcl/source/fontsubset/cff.cxx:637.

`nInt = static_cast<int>(nVal);`

This patch change casting from float to double.

I will apply this patch to the latest AOO418 and inform the result to this thread.

Pilot-Pirx:
Thanks for testing!

Can you have a look at PR #89?

This is an enhancement to the fix in 418-RC2.

ardovm:
Hello,

I am chiming is as the proposer of PR #89.
Unfortunately, we are still affected by the bug. I am investigating.

ardovm:
@androidports 
> I think 4.1.8-RC2 avoid crashing by invalid CFF format, but this invalid format is caused by casting from `float` to `int` in main/vcl/source/fontsubset/cff.cxx:637.
> 
> `nInt = static_cast<int>(nVal);`

Very good catch, Yoshikawa-san!
A `float` cannot represent all values of a 32-bit `int`!

I wonder why no one ever found this bug before?

> This patch change casting from float to double.

I confirm this fixes the garbled PDF output.

IMHO this PR addresses a totally different problem than PR #89 and is worth merging on its own.

Pilot-Pirx:
I can replicate the garbled PDF output on Windows and Ubuntu.
Will now build 418 for Windows with the patch applied.

Pilot-Pirx:
Actually I build AOO42X, but I can confirm that the exported PDF looks OK now.
This will also go in AOO418-RC3.

Pilot-Pirx:
I almost forgot to say thank you for providing this patch! ;-)

androidports:
You are welcome!
I’m going to give feed backs from my project.

PR #109, opened 2020-11-21
Allow boost::tr1 to build with Gcc version >= 7

ardovm:
As already pointed out by [bug 127712](https://bz.apache.org/ooo/show_bug.cgi?id=127712), Gcc seems to have changed its behavior, with respect to filesystem paths, since Gcc 7. 

The Boost::tr1 libraries expect C++ standard library headers to be installed into paths that contain the full Gcc version (major.minor.patchlevel). This is may not happen since Gcc 7.

This PR is aimed to allow compilation of current AOO sources on systems with Gcc version 7 or later, and C++ headers installed accordingly to the more recent standard. It was tested on openSUSE Leap 15.2 x86_64.


Pilot-Pirx:
Did anyone have the time to have a look at this PR?

jimjag:
LGTM

Pilot-Pirx:
I will merge it then... ;-)

PR #111, opened 2020-12-08
Enable compilation on Linux systems putting libraries in /lib64 instead of /lib

ardovm:
Some Linux distributions, such as openSUSE, on x86_64 systems install libraries in directories named "lib64" instead of "lib".

AOO418 and development branches do not compile successfully, because the Python module follows the above convention, being the system default, but the OpenOffice build system does not expect that and does not deliver the `.so` files.

This PR has the purpose to restore compatibility with such Linux distributions.

The only proposed commit so far is for the Python module. I tested it on my openSUSE Leap 15.2 x86_64 system.

In order to be sure that no other module is showing problems, I would like to try replicating an "official release" build, but I could not find any information about it on the Wiki. Pointers would be appreciated.


Pilot-Pirx:
I cannot imagine any harm this PR could do.
If no one objects I will merge it later in the afternoon.

PR #112, opened 2020-12-15
updated handling of repeating groupheaders

Grosskopf:
A fix for [Bug 108383](https://bz.apache.org/ooo/show_bug.cgi?id=108383)

Repeating groupheaders were in the page's header while normal groupheaders were in the pages body. So when a normal groupheader was before a repeating groupheader, it would go to body first, write the normal groupheader and then "return" to the next pages header to write the repeating groupheader since that is the next possible header to write to.

In this fix I have changed this, so that a repeating groupheader writes to the page body like the normal groupheader, but also creates a followup pagestyle that has the repeating groupheader in the header.

This was tested mostly in Libreoffice since their [bug 51452](https://bugs.documentfoundation.org/show_bug.cgi?id=51452) is the exact same and the source code is not so different in this area but it was tested once in open office and created similar results.

### DIFF
--- a/main/reportbuilder/java/com/sun/star/report/pentaho/model/OfficeMasterPage.java
+++ b/main/reportbuilder/java/com/sun/star/report/pentaho/model/OfficeMasterPage.java
@@ -59,4 +59,14 @@ public class OfficeMasterPage extends Section
     {
         setAttribute(OfficeNamespaces.STYLE_NS, "page-layout-name", name);
     }
+
+    public void setNextMasterPage(final String name)
+    {
+        setAttribute(OfficeNamespaces.STYLE_NS, "next-style-name", name);
+    }
+
+    public String getNextMasterPage()
+    {
+        return (String) getAttribute(OfficeNamespaces.STYLE_NS, "next-style-name");
+    }
 }
--- a/main/reportbuilder/java/com/sun/star/report/pentaho/output/OfficeDocumentReportTarget.java
+++ b/main/reportbuilder/java/com/sun/star/report/pentaho/output/OfficeDocumentReportTarget.java
@@ -741,10 +741,22 @@ public abstract class OfficeDocumentReportTarget extends AbstractReportTarget
         return groupContext;
     }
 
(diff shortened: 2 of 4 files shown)


leginee:
Awesome! Thanks for the patch.

While reviewing the patch, I noticed that some code has only commented out and not removed. Is there any reason why we want to keep it?

see main/reportbuilder/java/com/sun/star/report/pentaho/output/text/TextRawReportTarget.java (patched line: 1245 to 1248)
`1245        /*if (getGroupContext().isGroupWithRepeatingSection()) in to  `

Grosskopf:
Thank you for the quick answer :) True, we don't need that. I tried Squashing it but that didn't quite work the way I expected xD is this ok?

Also, I committed this patch to Libreoffice too (https://gerrit.libreoffice.org/c/core/+/107780/3), so it would be under Apache License for you and under multiple licenses for TDF right? :)

leginee:
If not needed I would would opt that the dead code is removed. :) Can you remove it? or should I check how I can incorporate this wish?

We need only the Apache License. If I were you I would add the License to the Comments that the patch is under dual License (Apache License, Mozilla License). But this is me.

Grosskopf:
> If not needed I would would opt that the dead code is removed. :) Can you remove it? or should I check how I can incorporate this wish?

I think I did remove it, is it not gone? this commit history is messy...

> We need only the Apache License. If I were you I would add the License to the Comments that the patch is under dual License (Apache License, Mozilla License). But this is me.

should I do this in the code? in the conversations around the patches I mentioned this...

Also I just found bugs I propably did in this patch... I tested another report without repeating groupheaders and it's not working anymore, not sure why... so this shouldn't be merged as of now...

leginee:
> I think I did remove it, is it not gone? this commit history is messy...

Probably I had some caching issue or tomatos on my eyes. I did not see that you removed the code, but now it looks good.
> Also I just found bugs I propably did in this patch...

Okay. No issue. We have time.

PR #113, opened 2020-12-16
Avoid "dmake clean" to wipe out the whole tree

ardovm:
As per a recent thread on the dev@ mailing list, the "dmake clean" command may wipe out the whole source tree, if the build environment was not set by sourcing the appropriate `.Env` script.

This PR has the purpose to fix the problem, assuming that if the build environment is not set, then the "dmake clean" command must not do anything.

### DIFF
--- a/main/makefile.rc
+++ b/main/makefile.rc
@@ -59,6 +59,7 @@ distclean .PHONY: clean
 
 
 clean .PHONY:
+	@test -n "$(INPATH)" || (echo Build environment not set; exit 1)
 	-rm -rf */$(INPATH)
 	-rm -rf solver/*/$(INPATH)
 .IF "$(ADDITIONAL_REPOSITORIES)"!=""


leginee:
hmm, that echo needs no quotes in 
` echo Build environment not set; works without Quotes?` 
looks strange. Is it special on make files that they can omit these?

ardovm:
@leginee I was also surprised it works :-)

Proof:
```
$ echo Hello world; echo Hallo Welt
Hello world
Hallo Welt
$
```

ardovm:
By the way, I followed that "pattern" as is it the same for the `install` target in the same file:

```Makefile
install .PHONY :
	@test "$(DESTDIR)$(prefix)" != "" || (echo Usage: make install [DESTDIR=DIR] prefix=DIR; exit 2)
[...]
```

leginee:
Worth a test. :)

cbmarcum (approved):
This fixed it for me.  Thanks!!

Pilot-Pirx:
Is there any reason not to merge this PR?

PR #115, opened 2021-01-03
Flaky tests

cbmarcum:
This work is for flaky BVT tests that randomly pass or fail during multiple runs against the same build.
Basically I'm adding a thread sleep prior to the test assertion check.
So far this has reduced the average failure count of 2 or 3 per run to 0 for the last 5 runs I've made against trunk.

### DIFF
--- a/test/testgui/source/bvt/gui/BasicFunctionTest.java
+++ b/test/testgui/source/bvt/gui/BasicFunctionTest.java
@@ -712,6 +712,7 @@ public class BasicFunctionTest {
 		scFunctionWizardDlgFunctionList.select("ABS");
 		scFunctionWizardDlgNext.click(); // Use "Next" button
 		scFunctionWizardDlgEdit1.typeKeys("A1");
+		sleep(1);
 		scFunctionWizardDlg.ok();
 		sleep(1);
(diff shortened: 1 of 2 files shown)


leginee:
I checked through the code and for me it is fine. I have a question thought.
The assert call had always a parameter as a implicit nested Function. and you broke that up, by declaring a variable first and then hand it to the assert function. Did you do this for style reasons or was there a deeper technical thought behind it?
Also note that the sleep comes always directly in front of the assert, so there is a change. That is why I wonder a bit.
I like the structure, because it is cleaner now.

cbmarcum:
The reason was technical.  My belief is the tests were passing and failing randomly either from threading issues or processor speeds that didn't exist when they were created.  When the tests would fail the error was the calculated values in formulas or grabbing from the clipboard was returning to slow to make it in time for the evaluation in the assertion.  By splitting them out I was able to add a sleep in thread before the assertion and this seemed to improve the reliability of the tests.

leginee:
Hmm, can I add a review confirmetion somehow, I wonder.

leginee (commented):
The Code looks oklay for me. Build bots have build the Source without issue.
Looks Okay to me.

PR #117, opened 2021-01-09
Changed helptext after resolved problem.

DiGro:
[:space:] should represent any whitespace character in regular expressions.
However, in the older, before OOo 3.4, it did not find the tab \t.

Now it does due to changing the regexp engine since 3.4


ardovm (approved):
This looks good to me. I propose merging it in 24 hours if no one raises any objections.

Pilot-Pirx (approved):
Looks OK for me!

PR #118, opened 2021-01-11
Tentative solution of bug #125129

ardovm:
See https://bz.apache.org/ooo/show_bug.cgi?id=125129


Pilot-Pirx (approved):
I have now done a Test build for Windows with this patch applied and can confirm that the crash is gone.
I cannot comment on the code itself, but my impression is that it is unlikely to break anything.

Pilot-Pirx:
If there are no objections I will merge this code later this day.

leginee (approved):
I agree, impact should be low.
The Function ist Titled by the way with "//Rahmen neu anlegen."
It should say // create new frame.

PR #124, opened 2021-03-10
Delete .hgtags

Pilot-Pirx:
Is this still needed?


Pilot-Pirx:
The same for .hgignore.
And do we really need more than one .gitignore?

cbmarcum (approved):
since these are from the old Mercurial system, I don't see why we need them.

PR #126, opened 2021-04-14
Fix to get the version string right in the About dialog

Pilot-Pirx:
This is not really a good solution. But we can use it until someone else has a better idea.

### DIFF
--- a/main/cui/source/dialogs/about.cxx
+++ b/main/cui/source/dialogs/about.cxx
@@ -502,14 +502,14 @@ const rtl::OUString AboutDialog::GetBuildId() const
     // For better readability, strip the duplicate ProductMajor ("340").
     if ( sProductSource.getLength() )
     {
-        sal_Int32 versionLength = sProductSource.getLength() - 3;
+        sal_Int32 nMajorLength = sProductSource.getLength() - 3;
         bool bMatchingUPD =
                 ( sProductSource.getLength() >= 3 )
-            &&  ( sBuildId.getLength() >= versionLength )
-            &&  ( sProductSource.copy( 3 ) == sBuildId.copy( 0, versionLength ) );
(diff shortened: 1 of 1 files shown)


ardovm:
I apologize for the forced push. I always have to cherry-pick two commits from trunk in order to build AOO41X on openSUSE and I included them by mistake

Pilot-Pirx:
Looks good!

ardovm:
This code is an attempt to glue two strings together, avoiding to repeat the common part in between. In order not to make a generic algorithm, that would be overkill (such as trying with one character, then with two, then with three etc.) we have to make some assumptions. The assumption is that the product source is always "AOOsomething" -- i.e. the "something" part starts at the fourth character.

I tried to find the definition of the "product source", but after passing from C++ to XML, to Perl, I got lost in the configure script and surrendered.

main/cui/source/dialogs/about.cxx
```diff
@@ -502,13 +502,14 @@ const rtl::OUString AboutDialog::GetBuildId() const
     // For better readability, strip the duplicate ProductMajor ("340").
     if ( sProductSource.getLength() )
     {
+        sal_Int32 versionLength = sProductSource.getLength() - 3;
```
ardovm: Actually, this should be named "productMajorLength" to be consistent with the above comment. It's probably worth renaming for the sake of clarity.

ardovm (approved):
It looks good now. I tried with with very long and very short versions and it seems to work.

I suggest we squash the commits when merging to avoid polluting the commit log.

Pilot-Pirx:
> I suggest we squash the commits when merging to avoid polluting the commit log.

Please go ahead, it is your code! ;-) Squash and merge will be best.

jimjag:
Approved by me

PR #127, opened 2021-04-15
Allow all https(s) Hyperlinks

dave2wave:
There was some debate on which is best for 4.1.10. Here is the PR should we go this way.

### DIFF
--- a/main/sfx2/source/appl/appopen.cxx
+++ b/main/sfx2/source/appl/appopen.cxx
@@ -944,10 +944,11 @@ void SfxApplication::OpenDocExec_Impl( SfxRequest& rReq )
                 // We consider some protocols unsafe
                 sal_Bool bUnsafeProtocol;
                 switch (aINetProtocol) {
-                // case INET_PROT_FTP:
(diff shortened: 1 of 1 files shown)


jimjag:
+1

cbmarcum (approved):
I can confirm the change now allows https url links to files with any extension, not just those on the whitelist to be passed on to the browser to handle.  Do we need to discuss on dev@ whether we should warn against these as suggested or not before merging?

dave2wave:
Well the CVE was about non-http(s) links. I think we are good to go. Especially since no one engaged to reply.

cbmarcum:
+1

cbmarcum:
Just to be clear this also allows any extension with HTTP also.  Just wanted to make sure that was the intent.

dave2wave:
Yes. That was the intent. Maybe in 4.2.0 we discussing forcing http to https as one of our hyperlink security settings.

cbmarcum:
I figured so but I thought I should check.  I've seen people get surprised by the fall though behavior in a switch case and I didn't check for HTTP before.

PR #132, opened 2021-06-12
Bug128453

ardovm:
This PR aims to solve bug 128453: a new security warning is being displayed also when it is not necessary.

This PR is against branch AOO41X because I started from there. After we merge it I will cherry-pick its commits into trunk and AOO42X.

### DIFF
--- a/main/officecfg/registry/data/org/openoffice/Office/Security.xcu
+++ b/main/officecfg/registry/data/org/openoffice/Office/Security.xcu
@@ -469,6 +469,276 @@
 				<value>xhp</value>
 			</prop>
 		</node>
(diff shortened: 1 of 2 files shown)


cbmarcum (approved):
I have built and tested this PR. I manually tested the same files I shared on the dev list [1].
I also ran the automated tests and all the hyperlink tests passed except the one I now need to remove due to the behavior of AOO when adding a host only smb: file link and it adding a trailing "/" which then doesn't throw the dialog.

[1] https://home.apache.org/~cmarcum/hyperlink-tests/

cbmarcum:
Based on the positive test results on the dev@ list I'm going to merge this in.

PR #133, opened 2021-06-19
graphite: remove the -Werror option from debug builds

ardovm:
The flag is set by default by the graphite configure script, but it lets the compilation fail with gcc 7.5

The compilation of the graphite module is consistently failing on openSUSE Leap 15.2 x86_64, when AOO is being built with debug options enabled. This patch allows successful compilation of the graphite module on such systems.

### DIFF
--- a/main/graphite/graphite-2.3.1_debug.patch
+++ b/main/graphite/graphite-2.3.1_debug.patch
@@ -11,3 +11,14 @@
  														// in the case of a non-contiguous list
  	return m_pseg->m_prgginf[(*m_vit) - m_pseg->m_isloutGinf0];
  }
+--- misc/silgraphite-2.3.1/engine/configure	2021-06-19 15:17:20.741894267 +0200
++++ misc/build/silgraphite-2.3.1/engine/configure	2021-06-19 15:18:16.359429854 +0200
+@@ -16877,7 +16877,7 @@
+ # (note: the flags here are gcc-specific and may fail with other compilers)
+ build_flags=""
+ if test "$enable_debug" = yes; then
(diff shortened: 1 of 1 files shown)


leginee:
The meaing of the Flag is
-Werror: every warning is treated as an error.

Since Grafite is a dependency it is fine to build without. However Imho we should strife for removing Warnings from Code.
Maybe it is worth trying to build with a new Version of Grafite on Suse Machines?

DonLewisFreeBSD:
> The meaing of the Flag is
> -Werror: every warning is treated as an error.
> 
> Since Grafite is a dependency it is fine to build without. However Imho we should strife for removing Warnings from Code.

Yes.  It could be an indication of an actual bug.

> Maybe it is worth trying to build with a new Version of Grafite on Suse Machines?

It is long dead upstream.  The replacement has an incompatible API.

leginee:
Then we have 2 choices. Fix the Issues or prepare the switch to a replacement.

ardovm:
The first two warnings are IMHO bugs. The first one is a function parameter type `bool` instead of `bool&` and the second is a mixup of return values in a `switch` statement.

Others are linked to assertions not being compiled (although we are requesting a debug build). Some others are unused variables that are explicitly marked `TODO` in the code.

I stopped after checking the first four warnings.

I suggest we allow this PR through so that at least the source becomes compilable again. Then we could have a longer term plan for the replacement of the graphite module with... whatever is better today.

For what it's worth, both the [SourceForge project page](http://silgraphite.sourceforge.net) and [www.sil.org](https://software.sil.org/products/#supported) refer to http://graphite.sil.org that does not seem to exist.

leginee:
Okay, lets merge. The issues are there anyhow. I dont think the code is better or worse with the change.

PR #134, opened 2021-06-25
piecewise enabling of category b components

cbmarcum:
Patch by: Yury Tarasievich

### DIFF
--- a/main/configure.ac
+++ b/main/configure.ac
@@ -1428,6 +1428,62 @@ dnl ===================================================================
 dnl Build options
 dnl ===================================================================
 
+# !!!
+AC_MSG_CHECKING([whether to enable CoinMP explicitly])
+enable_coinmp_explicit=
+if test -n "$enable_coinmp" && test "$enable_coinmp" != "no"; then
+   enable_coinmp_explicit="yes"
+   AC_MSG_RESULT([yes])
+else
+   enable_coinmp_explicit="no"
+   AC_MSG_RESULT([no])
+fi
+AC_MSG_CHECKING([whether to enable Graphite explicitly])
+enable_graphite_explicit=
+if test -n "$enable_graphite" && test "$enable_graphite" != "no"; then
+   enable_graphite_explicit="yes"
+   AC_MSG_RESULT([yes])
+else
+   enable_graphite_explicit="no"
+   AC_MSG_RESULT([no])
+fi
+AC_MSG_CHECKING([whether to enable Saxon explicitly])
+enable_saxon_explicit=
+if test -n "$enable_saxon" && test "$enable_saxon" != "no"; then
+   enable_saxon_explicit="yes"
+   AC_MSG_RESULT([yes])
+else
+   enable_saxon_explicit="no"
+   AC_MSG_RESULT([no])
+fi
+AC_MSG_CHECKING([whether to enable Rhino explicitly])
+enable_javascript_explicit=
+if test -n "$enable_javascript" && test "$enable_javascript" != "no"; then
+   enable_javascript_explicit="yes"
+   AC_MSG_RESULT([yes])
+else
+   enable_javascript_explicit="no"
+   AC_MSG_RESULT([no])
+fi
+AC_MSG_CHECKING([whether to enable Hyphenator explicitly])
+enable_hyphen_explicit=
+if test -n "$enable_hyphen" && test "$enable_hyphen" != "no"; then
+   enable_hyphen_explicit="yes"
+   AC_MSG_RESULT([yes])
+else
+   enable_hyphen_explicit="no"
+   AC_MSG_RESULT([no])
+fi
+AC_MSG_CHECKING([whether to enable Hunspell explicitly])
+enable_hunspell_explicit=
+if test -n "$enable_hunspell" && test "$enable_hunspell" != "no"; then
+   enable_hunspell_explicit="yes"
+   AC_MSG_RESULT([yes])
+else
+   enable_hunspell_explicit="no"
+   AC_MSG_RESULT([no])
+fi
+
 # Disable or enable category-B libraries: nss, hunspell, hyphen, saxon,
 # rhino, silgraphite.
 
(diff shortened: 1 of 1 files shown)


yurytch:
Re: https://www.mail-archive.com/dev@openoffice.apache.org/msg42099.html
Part of the complete set of options enabling building only 'hyphenation' and 'spelling' modules from 'category B license' set:
```
--disable-category-b \
  --enable-hyphen \
  --enable-hunspell \
```

ardovm:
It seems to me that when indicating  `--enable-category-b` all the new choices about Graphite, Hunspell etc. default to "no". It should be "yes" instead, if we want to keep the previous behavior.

main/configure.ac
```diff
@@ -4392,7 +4449,7 @@ _ACEOF
           cat saxontest.java >&5
           AC_MSG_ERROR([saxontest could not be compiled, non-functional saxon jar])
         fi
-elif test "x$enable_saxon" = "xyes"; then
+elif test "x$enable_saxon" = "xyes" -a test "x$enable_saxon_explicit" != "xno"; then
```
ardovm, line 4452: Instead of `(enable_saxon = yes) -a (enable_saxon_explicit != no)` it should rather be `(enable_saxon = yes) -a (disable_saxon_explicit != no)`

In this patch, not indicating explicitly one setting is considered as disabling it explicitly.

The previous implicit behavior is not preserved.

cbmarcum: Thanks Arrigo for the review.  I will update per your suggestion.

yurytch: > In this patch, not indicating explicitly one setting is considered as disabling it explicitly.

This patch deals only partially with the behaviour of `--enable-category-b` option, namely that specifying `--disable-category-b` voids any subsequent selective enabling of the category-b parts.
Therefore there are indeed more of such modifications in the `configure.ac` to make, for someone more knowledgeable than me. In my local copy I also have `saxon` explicitly selectable, but also `graphite`, `javascript`, and `coinmp`. There are more `category-b` parts in the `configure.ac`, if I remember right.

cbmarcum: @Yurytch are you able to update this PR per Arrigo's comments?

yurytch: > @yurytch are you able to update this PR per Arrigo's comments?

The change in the Arrigo's comment looks wrong to me.
My patch deals with ONE specific problem, how to enable category-b components piece by piece, even if category-b itself is disabled (the previous behaviour was that enable directives for category-b components have no effect in this case).
So I added those additional configure directives, kind of 'on top' of the existing AC code.
Arrigo talks about something else, but his change does not seem to address HIS issue, it tweaks MY solution, and it looks that it breaks it.

Maybe I just don't understand the issue well enough.

ardovm: > > @yurytch are you able to update this PR per Arrigo's comments?
> 
> The change in the Arrigo's comment looks wrong to me.

Yes, it was not meant to be applied verbatim. It was just a suggestion about how variables could be named.
I am sorry if I led you to think otherwise.

> My patch deals with ONE specific problem, how to enable category-b components piece by piece, even if category-b itself is disabled (the previous behaviour was that enable directives for category-b components have no effect in this case).
> So I added those additional configure directives, kind of 'on top' of the existing AC code.
> Arrigo talks about something else, but his change does not seem to address HIS issue, it tweaks MY solution, and it looks that it breaks it.
> 
> Maybe I just don't understand the issue well enough.

I wanted to point out that your patch changes the current behavior when the only `--enable-category-b` parameter is given.

In my humble opinion, your patch should add choices, but not change the behavior of currently available parameters. So instead of disabling all "category b" parameters by default, you could rather enable them, and add switches to disable each of them explicitly. Such switches could set variables named `disable_xxx_explicit`.

I hope it is more clear now.

yurytch: If I (finally) get you right, you are saying that my patch breaks `--enable-category-b` behaviour. 
My apologies.
However, the fix should go like that, then:

`+if \( test -n "$enable_saxon" && test "$enable_saxon" != "no" \) -o "$enable_category_b" = "yes" ; then`

for every equivalent string in the 1st part of the patch.

ardovm: IMHO a variable that is named `enable_saxon_explicit` should be set when we "explicitly" indicate `--enable-saxon`.
I suggested to use another variable `disable_saxon_explicit`, that should be set when we "explicitly" indicate `--disable-saxon` but it would also work to use `enable_saxon_explicit="no"`, as you suggest.

If we do not either enable or disable something explicitly, then we may enable it "implicitly" by means of `--enable-category-b` or disable it implicitly by means of `--disable-category-b`.

When it's time to understand if saxon has to be enabled or not, the logic should be:
1. was it either enabled or disabled explicitly? If so, enable or disable it.
2. was it enabled implicitly? If so, enable it.
3. Otherwise disable it.

With your last comment, you seem to be getting close to this, but I do not understand what you mean by "explicit".
I hope I could explain well what _I_ would mean. I hope it makes sense to you.

yurytch: Arrigo, you seem to miss (or skip over) the way `--enable-saxon` option actually works (and likewise its counterparts for other category-b members).
The option does NOT work i.e. has **no effect** if `--disable-category-b` option is also given.

So one can't disable category-b components as a group and still enable separate components.
I didn't want to touch that part of configure code, so I added `enable_saxon_explicit` variable to work around the problem. 
I sort of strengthened `--enable-saxon`'s effect (likewise for its counterparts for other category-b members).

If it broke selective DISabling saxon when enable-category-b is also given, then we should look at how to strengthen the disabling part of the code -- because your patch looks like it effectively brings back the original AC behaviour, before MY patch.

ardovm: I am sorry this discussion is taking so long.

As I wrote initially, this pull request changes the current behavior of the configure script, and IMHO this should not happen.

I am not proposing any patches. I am just trying to help by reviewing yours. If you will want to propose an amended version, I will be happy to review it again.

yurytch:
I didn't know the current behavior of the configure script should not be changed.
If that's so, let's give the issue a big goodbye. 
I'll propose an amended version some time later.

cbmarcum:
Closing per @yurytch

PR #139, opened 2021-10-03
Occurence "OpenOffice.org" in dialog Writing Aids for new thesaurus

DiGro:
Just noticed the occurrence of the string "OpenOffice.org New Thesaurus" in the
dialog Writing Aids (Tools > Options > Language Settings > Writing Aids) in the
Developer Build 4.2.0

This should probably be changed to "New Thesaurus" only.

OpenGrok found the string just once, in:

/trunk/main/lingucomponent/source/thesaurus/libnth/nthesimp.cxx (revision
c4c42a0)

Line:
576 {
577 MutexGuard aGuard( GetLinguMutex() );
578 return A2OU( "OpenOffice.org New Thesaurus" );
579 }

Since I'm no developer at all I just deleted "OpenOffice.org".
Hope someone will check if I did it as supposed to.

Should probably also need to be changed in branches AOO41 and AOO42 too

Earlier reported in BugZilla Issue ID: 128483 https://bz.apache.org/ooo/show_bug.cgi?id=12
Options_Language settings_Writing aids
84

### DIFF
--- a/main/lingucomponent/source/thesaurus/libnth/nthesimp.cxx
+++ b/main/lingucomponent/source/thesaurus/libnth/nthesimp.cxx
@@ -575,7 +575,7 @@ OUString SAL_CALL Thesaurus::getServiceDisplayName( const Locale& /*rLocale*/ )
 		throw(RuntimeException)
 {
 	MutexGuard	aGuard( GetLinguMutex() );
-	return A2OU( "OpenOffice.org New Thesaurus" );
+	return A2OU( "New Thesaurus" );
 }
 
 


ardovm:
Looks good to me, because the string is not translated.

We could argue that translating this and other strings returned by `getServiceDisplayName(const Locale&)` methods would be a good idea: the `Locale` parameter is there for a reason... but IMHO it is not the purpose of this PR.

ardovm (approved):

dave2wave:
This should be merged to AOO42X and AOO411X as well as trunk

Pilot-Pirx:
+1 for merging in all branches.

I am not sure if we should keep the "OpenOffice" for historical reasons but I am OK with "New Thesaurus"

PR #150, opened 2022-05-29
Refs #i112383#, #i117960#, and #i126272# - Basic lang backports

cbmarcum:
These are a few fixes for the Basic compiler that are in trunk and AOO42X but not backported to AOO41X.

Can't cherry-pick due to file changes and two commits changed the same test file.
Each of these are covered by a test now that I've added one for i126272.

Descriptions and original commit messages and links follow:
Fix for error in Convert to Long function.
original commit:
#i112383# CLng('&HFFFFFFFF') fails on 64-bits rather than returning -1
Found-by: andrew
Patch-by: Damjan Jovanovic
https://github.com/apache/openoffice/commit/323c350123d05cd9f450aebd421b7bfa74c331d2

Fix for using certain keywords as variable names in single line if statements
Original commit:
#i117960# Basic: Line Input doesn't work in single-line If
i92642 added the ability to use certain keywords as variable names (eg. name = 1, line = 'hi'),
but also caused a regression where "Line Input" is broken in single-line If statements.
This patch fixes that by allowing Then and Else to also be the start-of-line tokens expected to
immediately preceed the 'Line' token in order for that 'Line' token to be recognized a keyword instead
of a variable name. Also added FVT spreadsheet tests for 'Line' as both a variable name and as 'Line Input'.
Patch by: Damjan Jovanovic
https://github.com/apache/openoffice/commit/725d867363770a4c47c4b3c6dbe1f359c296604a

Basic compile error when comment is at the end of a single line if then else statement
original commit:
#i126272# OpenOffice.org Basic compile error : if statement followed by End If - in next Line ???
Allow the Else in a single-line If statement to be terminated by a comment instead of only EOL.
Patch by: Damjan Jovanovic
https://github.com/apache/openoffice/commit/07396187f6055b1e7cffa86f38cc88b274dfb1d6


cbmarcum:
I found these changes missing from AOO41X after updating the tests in trunk to be standalone and running them against AOO41X and debugging test failures and errors. To run the included test after building this branch. 
From test dir:
```
ant -Dopenoffice.home="<path-to-openoffice4>" compile
./run -Dopenoffice.home="<path-to-openoffice4>" -tc fvt.uno.sc.formula.TestFormulaDocs
```
The test in trunk has been further improved as a JUnit parameterized test so each document is reported separately.

cbmarcum:
Based on this email thread [1] on dev@ I'm going to pull this in.

[1] https://lists.apache.org/thread/21vgh6ovcp7mnf82j7nz6rngzx7yvztn

PR #151, opened 2022-06-09
Fix for Issue 114497

oooforum:
Thanks to hanya for this patch


cbmarcum:
Hi Cyril, could you provide some simple steps we can use to test this before and after the fix for those who have not used the mail merge feature before.

oooforum:
Simple, read the BZ report https://bz.apache.org/ooo/show_bug.cgi?id=114497 and apply instructions from comment #37
Before, we can not use smtp with SSL protocol (like GMail account).
After, access works.

oooforum:
Somebody can push it this fix? 
It will be fine to have this in the next builds.

ardovm (approved):
The code looks good to me.

PR #153, opened 2022-06-13
Fix for Issue 128375

leginee:
This is a pull request for[ Issue 128375](https://bz.apache.org/ooo/show_bug.cgi#add_comment)

I removed all ~ in Text Aerea.
Example:
```
String STR_EDIT_OBJ
{
	Text [ en-US ] = "~Edit" ;
};
```
or reported
```
`String STR_PLAY
{
	Text [ en-US ] = "~Play" ;
};`
```
Not sure maybe there is an impact on other language files?


cbmarcum:
Are you sure the tilde doesn't have some significance like in hotkeys?
Also, which file has this change.  I see removing stax and some some build files?

leginee:
> Are you sure the tilde doesn't have some significance like in hotkeys? 

Interesting point. Needs to be reviewe?. Then why is Tilde visible in the string as reported.

> Also, which file has this change. I see removing stax and some some build files?

oO. There went something wrong. I need to check. I withdraw my PR then.

PR #154, opened 2022-08-30
Update srcrelease.xml to exclude certain files from Source build

Pilot-Pirx:


Pilot-Pirx:
See:
https://bz.apache.org/ooo/show_bug.cgi?id=127544

ardovm (changes requested):
You are excluding all files named `.log`. You probably wanted to exclude all files _ending_ in `.log`? If so, the correct entry should be `**/*.log`

BTW, well spotted the missing comma after `**.svn`!

ardovm:
It is also not very clear to me, if the other spurious files from the bug report are still included.
If so, shall we also add exclusions for them in this PR?

Pilot-Pirx:
Thanks! My understanding is that we don't want to bundle any log, so I will change it accordingly.

ardovm (approved):
This looks good to me.
Any other files to exclude?

Pilot-Pirx:
I think Andrea had some...
If he can add them we could close the Bugzilla issue.

Pilot-Pirx:
I think this issue should be fixed before the next release.
So any additions are welcome!

Pilot-Pirx:
If there are no additions needed I will merge this PR in the next days.
Then it will be included in 4.2.0-Dev4 and 4.1.14.

PR #160, opened 2022-11-12
Remove SuSE packaging information

ardovm:
The SuSE RPMs depend on certain files that were once present in SuSE distributions, and were deprecated.

Freedesktop RPMs should work flawlessly on today's SUSE and openSUSE distributions.

See https://bz.apache.org/ooo/show_bug.cgi?id=128424


ardovm:
I will merge this PR in one week if no one objects

PR #161, opened 2022-11-20
Increase defaults for graphic cache

Pilot-Pirx:
As discussed on dev@:
 - maximum cache size for all graphical display objects: 100 MB
 - maximum cache size for a single graphic display object: 10 MB


marcus-aoo:
You habe found the location in the code, great. :-)

Pilot-Pirx:
However, this will only have an effect on new installations, I think.

Pilot-Pirx:
If there are no objections I will merge this PR at the end of the week to trunk and cherry-pick it for AOO42X. This way it will be in 4.2.0-Dev4 for testing.

PR #165, opened 2023-01-12
Update zlib to 1.2.13

Pilot-Pirx:
Currently we use zlib 1.2.7 from 2012.
This PR updates it to the most recent version 1.2.13 from 2022.

I am no developer, but basically we only add a makefile and patch some headers?

I could build successfully with this PR on Windows, testing on other OS is welcome!


ardovm (approved):
Looks good to me! It builds fine under Linux.

Strictly speaking, changing permissions of `main/zlib/make_patched_header.pl` was not necessary IMHO, but it is ok.

Pilot-Pirx:
Thanks for the review!
The permission change was done by git, so I will revert it.

And I *think* on Linux and macOS we probably use system libs, so zlib will only be built on Windows?

PR #166, opened 2023-02-06
Translate swhtml

leginee:
I have translated all comments. Please check if everything makes sense. I used deepl.com as support. I hope all translation errors are fixed.


leginee:
this went wrong i try again

PR #171, opened 2023-02-19
Address bug #126680

ardovm:
The purpose of this branch is to fix bug #126680.
All information is on the bug report:
https://bz.apache.org/ooo/show_bug.cgi?id=126680

Instances of com.sun.star.i18n.Transliteration.l10n are cached and sometimes (!!) treated as singletons.
But they may be modified by the callers. This means that one caller may see its instance unexpectedly modified by another caller.

It _may_ be possible to keep some kind of caching, like for variable `Reference< XSingleServiceFactory > xFactory`. As I am not sure this would work, I am just removing any caching. In fact, it has to be demonstrated how many times this code is called, throughout a normal usage session of AOO.

### DIFF
--- a/main/i18npool/inc/transliterationImpl.hxx
+++ b/main/i18npool/inc/transliterationImpl.hxx
@@ -121,15 +121,6 @@ private:
     com::sun::star::uno::Reference< XLocaleData > localedata;
     com::sun::star::uno::Reference< com::sun::star::i18n::XExtendedTransliteration > caseignore;
 
-    /** structure to cache the last transliteration body used. */
-    struct TransBody
-    {
-        ::osl::Mutex mutex;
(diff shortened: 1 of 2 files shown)


leginee:
I do not agree that caching makes sense. If behavior is switched the data needs to be adjusted too.
IMHO it is better to work with less dynamic Objects. instead instantiate a new object.
I would prefer policy based class in this case as a development goal.

Pilot-Pirx:
Please test! This could be a good addition to our next release.

I an building for Windows now.

https://www.dropbox.com/s/fdnul45ksc1lkh1/Apache_OpenOffice_4.5.0_Win_x86_install_en-US_bug126680.exe?dl=0

Pilot-Pirx:
Any news here?

PR #172, opened 2023-02-25
Update icons for OS/2

Pilot-Pirx:


Pilot-Pirx:
I would like to add Yuri Dario for testing, but somehow I am not able to...

Pilot-Pirx:
I don't expect problems here since these icons were created on ArcaOS.
While I am no icon artist they are now conform to other platforms.
If someone can do it better just commit! ;-)

PR #173, opened 2023-03-07
fgrep is deprecated and may be removed soon, use grep -F instead

Pilot-Pirx:
I am not a developer, so comments are welcome!


Pilot-Pirx:
The same is valid for egrep -> grep -E

ardovm (approved):
Looks good to me.
However, I cannot find any invocation of the `makedepn` command.

Pilot-Pirx:
http://opengrok.openoffice.org/xref/trunk/main/solenv/bin/srcrelease.xml?r=0f464f24#33

But here it is only excluded from the source build...
Maybe it is "dead" code?

PR #174, opened 2023-03-07
Remove use of pthread_mutexattr_setkind_np due to compile errors

20kdc:
This is intended as an informative notification, since I have no idea how to actually create a bugzilla account.

The code being removed here causes a compilation issue on Ubuntu Jammy Jellyfish. (undefined reference to `pthread_mutexattr_setkind_np`) The best I can tell, https://bugs.launchpad.net/ubuntu/+source/glibc/+bug/2003146 explains this somewhat.

This workaround likely hasn't been necessary for the past 20 years, if glibc "converted to git" repositories are any indication (see for instance https://github.com/lattera/glibc/blob/d82e4c7bb231c9e0f835bd46467563ac3b56cebe/linuxthreads/sysdeps/pthread/pthread.h ).

And indeed, anyone who would be attempting to run OpenOffice on such a system would likely encounter issues with the other varied libraries in use.

Unfortunately other compilation issues prevented me from performing a full test, but osl mutex tests pass at least.

### DIFF
--- a/main/sal/osl/unx/mutex.c
+++ b/main/sal/osl/unx/mutex.c
@@ -30,9 +30,11 @@
 #include <stdlib.h>
 
 #if defined LINUX /* bad hack */
+#if !defined(__USE_XOPEN2K8)
 int pthread_mutexattr_setkind_np(pthread_mutexattr_t *, int);
 #define pthread_mutexattr_settype pthread_mutexattr_setkind_np
 #define PTHREAD_MUTEX_RECURSIVE PTHREAD_MUTEX_RECURSIVE_NP
+#endif
 #endif
 
 /*


ardovm:
Thank you for you report!

> This is intended as an informative notification, since I have no idea how to actually create a bugzilla account.

You could follow the instructions at https://bz.apache.org/ooo/createaccount.cgi or write to the [dev list](https://openoffice.apache.org/mailing-lists.html#development-mailing-list-public) asking for help.

> The code being removed here causes a compilation issue on Ubuntu Jammy Jellyfish. (undefined reference to `pthread_mutexattr_setkind_np`) The best I can tell, https://bugs.launchpad.net/ubuntu/+source/glibc/+bug/2003146 explains this somewhat.

Using `nm` as described on that report, we can demonstrate that symbol `pthread_mutexattr_setkind_np` _is_ still present on other current mainstream distributions:
```
$ lsb_release -d ; nm /usr/lib/x86_64-linux-gnu/libpthread.so.0 | grep pthread_mutexattr_getkind_np
Description:    Debian GNU/Linux 11 (bullseye)
000000000000ca30 W pthread_mutexattr_getkind_np
```
and
```
$ lsb_release -d ; nm /usr/lib64/libpthread.so | grep pthread_mutexattr_getkind_np
Description:    openSUSE Leap 15.4
000000000000f5de W pthread_mutexattr_getkind_np
```

On the other hand, applying your patch to AOO41X makes it not compile any more on our (admittedly outdated) Linux reference build systems.
(I know you are proposing a patch to trunk, but IMHO it may be worth to look at the bigger picture, if we can)

> This workaround likely hasn't been necessary for the past 20 years, if glibc "converted to git" repositories are any indication (see for instance https://github.com/lattera/glibc/blob/d82e4c7bb231c9e0f835bd46467563ac3b56cebe/linuxthreads/sysdeps/pthread/pthread.h ).

As explained above, your problems seems to be Ubuntu-specific, at least for now.

IMHO we should look for other ways (like `#ifdefs`) to allow compilation on both systems with and without the deprecated symbol.

20kdc:
I've modified the commit to test for the presence of `PTHREAD_MUTEX_RECURSIVE_NP` instead of removing the code.

ardovm:
That's a good try, but unfortunately it does not work, because the `PTHREAD_MUTEX_xxx` constants are enum members, not macros.

Please check how they are defined inside your `/usr/include/pthread.h`. Maybe there is another macro guarding them? Here is an excerpt from our reference systems:
```c
/* Mutex types.  */
enum
{
  PTHREAD_MUTEX_TIMED_NP,
  PTHREAD_MUTEX_RECURSIVE_NP,
  PTHREAD_MUTEX_ERRORCHECK_NP,
  PTHREAD_MUTEX_ADAPTIVE_NP
#if defined __USE_UNIX98 || defined __USE_XOPEN2K8
  ,
  PTHREAD_MUTEX_NORMAL = PTHREAD_MUTEX_TIMED_NP,
  PTHREAD_MUTEX_RECURSIVE = PTHREAD_MUTEX_RECURSIVE_NP,
  PTHREAD_MUTEX_ERRORCHECK = PTHREAD_MUTEX_ERRORCHECK_NP,
  PTHREAD_MUTEX_DEFAULT = PTHREAD_MUTEX_NORMAL
#endif
#ifdef __USE_GNU
  /* For compatibility.  */
  , PTHREAD_MUTEX_FAST_NP = PTHREAD_MUTEX_TIMED_NP
#endif
};
```
According to the above, checking for the ``__USE_UNIX98`` macro could be an alternative solution.

20kdc:
Oddly, `__USE_UNIX98` is not defined here. `__USE_XOPEN2K8` however is.
Plus, `#if !defined(__USE_XOPEN2K8)` presumably ensures this won't deactivate until systems >= 2008, which is definitely long enough that any systems that define `__USE_UNIX98` but don't have `PTHREAD_MUTEX_RECURSIVE` won't get caught up in all this.
I'll revise the commit accordingly.

ardovm:
Now _this_ works well! Thank you!
Merging now.

20kdc:
Thank you also.

PR #176, opened 2023-03-12
Compatibility with GNU Make 4.4 (slideshow module)

ardovm:
Make 4.4 seems to ignore the parameters given for object slideshowimpl for static library "sldshw_s". We request the object file here too, so we indicate the parameters (again) here.


Pilot-Pirx:
If it works, commit it! ;-)
When we find a better solution later we can always adjust...

BTW: We only need this for trunk/AOO42X

PR #177, opened 2023-04-09
Remove unused(?) code

Pilot-Pirx:
Gluepoint is actually generated here:
https://github.com/apache/openoffice/blob/trunk/main/svx/source/sdr/primitive2d/sdrprimitivetools.cxx#L80

This part looks like a leftover for me.

"SdrGluePoint::Draw" is only to be found in this file and nowhere else: http://opengrok.openoffice.org/search?project=trunk&full=%22SdrGluePoint%3A%3ADraw%22

Any opinions?


Pilot-Pirx:
I am also curious why "SdrGluePoint::Mirror" is defined two times?

aelnosu (commented):
not sure that it is actually unused.😕

ardovm:
Once we delete the implementation of method `SdrGluePoint::Draw`, it is IMHO a good idea to delete its declaration as well. In this way, if any other piece of code refers to that method, it should not compile any more.

Let us now check that this branch is still compiling.

Pilot-Pirx:
Thanks, I will check!

Update: It builds fine! And the glue points in Draw are still there...

ardovm (approved):
Looks good to me, and compiles under Linux.

PR #178, opened 2023-04-23
first implementation of datedif

leginee:
Attached my implementation of datedif.
 
test is missing, but i think it builds
work in progress


oooforum:
Thanks for this patch.
Can it be pushed for the next build?

DamjanJovanovic:
Great work, thank you!

Can you also please add a description of this function to our help system, probably under main/helpcontent2/source/text/scalc/01/?

PR #181, opened 2023-08-10
Set the fontcolor always to black

Pilot-Pirx:
We should set the fontcolor to black by default.

Some themes on Linux use a slightly different color, for example Ubuntu uses a dark gray.

That drove me mad, because I thought my printer was broken. ;-)
At the moment I have set the fontcolor manually to black.

### DIFF
--- a/main/svtools/source/config/colorcfg.cxx
+++ b/main/svtools/source/config/colorcfg.cxx
@@ -1,5 +1,5 @@
 /**************************************************************
- * 
+ *
  * Licensed to the Apache Software Foundation (ASF) under one
  * or more contributor license agreements.  See the NOTICE file
  * distributed with this work for additional information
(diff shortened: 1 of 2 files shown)


Pilot-Pirx:
BTW: This can be set manually via "Tools - Options - Appearance - Font color"

Pilot-Pirx:
It looks like the color of the page border in Writer is linked to the font color.
So we end up with a black rectangle on black background in HighContrast mode.
I don't understand the logic behind that so I think this PR does not make sense until this is fixed/changed.

Pilot-Pirx:
Now the page border in Writer is independent of the fontcolor.
That means it is black in normal mode and white in High Contrast mode.

Pilot-Pirx:
Unfortunately there are some more items that use the fontcolor in High Contrast mode.
Investigating...

PR #184, opened 2023-11-16
misc: fix spelling

jbampton:


main/offapi/com/sun/star/awt/XListBox.idl
Pilot-Pirx, line 1: This is a historical typo in our API. I think we would need to address both versions and set the wrong one to "deprecated".

@ardovm What do you think?

Pilot-Pirx: https://www.openoffice.org/api/docs/common/ref/com/sun/star/awt/XListBox.html#isMutipleMode

ardovm: Being it a long standing API, IMHO we should be very very motivated in order to change it.
I agree that duplicating it is a feasible way to go, for the moment.

ardovm (changes requested):
Please do not substitute API names.

Pilot-Pirx (approved):
I think we are good now!

main/winaccessibility/source/UAccCOM/MAccessible.h
ardovm, line 1: I would normally be very reluctant to rename methods as we are doing here.
However, these methods do not seem to be called anywhere, so renaming them should not break anything.
We should rather understand why this code is here, given that it is not used.
But let's make one step at a time.

ardovm (approved):
Approved. Thanks!

PR #185, opened 2023-11-25
misc: fix spelling

jbampton:


test/testcommon/source/org/openoffice/test/vcl/client/CommandCaller.java
Pilot-Pirx, line 1: This is a typo in one of our Java classes, see:
http://opengrok.openoffice.org/search?project=trunk&full=%22protocal%22&defs=&refs=&path=&hist=&type=

I would not dare to change it...

PR #193, opened 2023-12-23
Some cleanups to current boost usage

pgiffuni:
Two commits here:
One updates the current boost package to include more headers.
One sets better defaults for sinc_pi a math operation from boost we use in a graphic filter.

They pave the way for a future boost update.


pgiffuni:
The commit log for [48eb146] is wrong: it is not an update yet, just a cleanup of the 1.55 packaging. Any way to fix that?

PR #195, opened 2023-12-23
Update main0108.xhp

Pilot-Pirx:
We still build with (invisible) Help instructions to purchase OpenOffice. This should be removed.


Pilot-Pirx:
Any opinions on the "Purchase" functionality?

Pilot-Pirx:
The code for the "Purchase-Wizard" seems to be long gone...
So the Help file does not make sense anymore.

PR #198, opened 2024-01-17
Allow compilation of AOO41X under Ubuntu Jammy (GCC 11)

ardovm:
The AOO41X branch does not seem to build on Ubuntu 22.04 "Jammy".

The purpose of this pull request is to gather and review the modifications that make AOO41X build again.

The edits mostly consist of:
1. disabling -Werror compilation flags, because newer GCC versions find more warning than previous ones;
2. fixing compilation where necessary as per [this thread of the dev@ mailing list](https://lists.apache.org/thread/gbfm2djykqk9ksqw22vobklf6d4lm0ly)

The edit to the boost module was cherry-picked from trunk.

This branch shall be tested on Ubuntu Jammy, but also on other platforms, in order to make sure that we are not breaking anything.


ardovm:
It was better to start from trunk.
Trunk now builds on Jammy and we will be able to cherry-pick the relevant commits into AOO41X.

PR #201, opened 2024-02-09
Remove code for unused bmpmaker

Pilot-Pirx:
See:
https://bz.apache.org/ooo/show_bug.cgi?id=83198


DamjanJovanovic (approved):
I like it. Some of the tools do seem useful though, eg. g2g can convert between image file formats, I wonder if they should be moved to a unit test or some example code instead?

Pilot-Pirx:
And I like the idea of preserving useful code...
Could we keep it in some sort of "internal attic" maybe in
https://svn.apache.org/repos/asf/openoffice/devtools/ ?

Pilot-Pirx:
My Windows build was successful.
I would like to merge this PR to trunk now and let the Linux buildbot do its job. ;-)

PR #194, opened 2023-12-23
Update boost to version 1.64 + minor enhancements

pgiffuni:
Add some headers to the existing 1.55 boost internal package.

Make cleaner usage of boost's sinc_pi functionality. There is no need to limit the policy.

PR #202, opened 2024-02-15
Update boost - Part 1

Pilot-Pirx:
This is a remake of the first part of Pedro Giffuni's original PR:
https://github.com/apache/openoffice/pull/194


Pilot-Pirx:
Original comments from Pedro:

**Minor cleanup to internal boost header packaging**

We normally don't install all the headers available in boost. There is an
interesting set of functionalities like multiprecision and odeint that while we
are not using, but would be interesting to have available.

This is done to ease future boost updates but also in the hopes of making it
possible to use this functionality in a near future.

No functional change expected.

Pilot-Pirx:
**Let boost promote to double the sinc_pi() operation.**

Modern processors should handle this without any trouble.

pgiffuni:
Yeah .. please merge this.
I have been trying the bigger update on a linux VM but I can't find what is wrong and in the meantime this is real progress.

Pilot-Pirx:
Since I am not a developer, I am still not sure if I should commit this PR.
But until now nobody complained and it did build on all platforms...

So here we go!

PR #205, opened 2024-02-22
docs: fix spelling in code comments

jbampton:


Pilot-Pirx (approved):
I am always astonished how many typos there are still left after all these years... ;-)

PR #206, opened 2024-03-29
Link pthread library when using OpenSSL on Linux

ardovm:
On my openSUSE system some libraries inside modules ucb, oox and possibly others did not link, because symbols from the pthread library could not be resolved.

This is an attempt to include the pthread library when the in-tree OpenSSL is used.

This patch restores the ability to compile AOO... but I am not 100% sure whether it is the correct way to implement it.

### DIFF
--- a/main/RepositoryExternal.mk
+++ b/main/RepositoryExternal.mk
@@ -390,6 +390,11 @@ $(call gb_LinkTarget_add_libs,$(1),\
 	-lsocket \
 )
 endif
+ifeq ($(OS),LINUX)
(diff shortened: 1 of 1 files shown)


DamjanJovanovic:
Oh right, on Linux pthreads is in a separate library, and internal OpenSSL is linked statically, so the binary linking to OpenSSL has to link to pthreads as well. Forgot about that case. Sorry.

Your patch looks good. Thank you.

DamjanJovanovic (approved):

PR #221, opened 2024-07-05
Update 05200100.xhp : par_id was disabled

DiGro:
par_id "par_id3154586"  was disabled by changing "id" to "nil".

Reverted back to field "id"


Pilot-Pirx:
I agree that this looks suspicious. It also is the only occurrence of "nil" in our Help files.
It was added with:
https://github.com/apache/openoffice/commit/c54bac3d2aecf93b2b5d687e650e52f319ac6383
(Maybe the id was from LO Help?)

For those who want to get more info on our Help system, please read:
https://www.openoffice.org/documentation/online_help/

dave2wave (approved):

PR #225, opened 2024-10-03
Update bridges to fix build on FreeBSD PowerPC64

lenoil98:
The bridges were updated to allow build on powerpc64 with default (clang) compiler.

### DIFF
--- a/main/bridges/Library_cpp_uno.mk
+++ b/main/bridges/Library_cpp_uno.mk
@@ -131,6 +131,10 @@ $(eval $(call gb_Library_add_exception_objects,$(COMNAME)_uno,\
         bridges/source/cpp_uno/gcc3_freebsd_powerpc64/uno2cpp \
 ))
 
+$(eval $(call gb_Library_add_asmobjects,$(COMNAME)_uno,\
+		bridges/source/cpp_uno/gcc3_freebsd_powerpc64/call \
+))
+
 $(eval $(call gb_LinkTarget_set_cxx_optimization, \
         bridges/source/cpp_uno/gcc3_freebsd_powerpc64/cpp2uno \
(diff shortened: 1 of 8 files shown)


dave2wave:
I think we need a review from a FreeBSD builder from downstream.

pedlino:
> I think we need a review from a FreeBSD builder from downstream.

I don't have FreeBSD or a PowerPC... Sorry!

ardovm:
> I think we need a review from a FreeBSD builder from downstream.

Yes! Possibly someone with a PowerPC...

In the meantime, we could check that this patch does not break compilation on other architectures.

@lenoil98 thank you for your contribution. Could you please state that it can be used under the Apache License, version 2.0?

ardovm:
We also need Apache license header on new files, please

lenoil98:
Updated files with Apache license header in new files.

ardovm:
...maybe @pgiffuni could help here?

DonLewisFreeBSD:
On  3 Nov, Arrigo Marchiori wrote:
> 
> ...maybe @pgiffuni could help here?
> 

Curtis Hamilton ***@***.***> filed this FreeBSD bug report and
recently attached PowerPC64 patches. I haven't compared them to the
patches in this MR.  

https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=251417

ardovm:
For what it's worth, I verified that this PR does not seem to affect Linux and Windows builds.

pgiffuni:
I don't have PowerPC either but Curtis Hamilton is pretty trustrworthy and has contributed changes for FreeBSD-PPC in the past.

pgiffuni (approved):

PR #233, opened 2024-12-27
msbuild.yml

ijunxyz123:
.github/workflows


dave2wave:
Thank you for this commit. Unfortunately this does not look like a valid GitHub action workflow file.
The scripting steps at the beginning must be included within a Job Step.

PR #191, opened 2023-12-01
feat(actions): add `pre-commit` framework with `codespell`

jbampton:
Official -> "Git hook scripts are useful for identifying simple issues before submission to code review. We run our hooks on every commit to automatically point out issues in code such as missing semicolons, trailing whitespace, and debug statements. By pointing these issues out before code review, this allows a code reviewer to focus on the architecture of a change while not wasting time with trivial style nitpicks."

https://pre-commit.com/

Using a `pre-commit` framework speeds up development as a lot of tests can be run on the local machine giving instant feedback. So we don't have to wait for the CI / GitHub actions to run to get feedback. The pre-commit automatically fixes some of the issues when you do git commit and if there are any issues the tests are marked as red failed. Then you will need to commit again so that all the tests pass green.

When pre-commit runs with GitHub Actions on the GitHub website the hooks/tests either pass or fail.

There are many more pre-commit checks listed here -> https://pre-commit.com/hooks.html

Lets get this PR merged and then I will look at adding more pre-commit tests 👍

This PR adds `codespell` to our pre-commit hooks.

The words in `codespell.txt` are ignored and this file has basically been created by running:

`codespell . | cut -f2 -d' ' | tr A-Z a-z | sort | uniq > codespell.txt`

from the repo root.

https://github.com/codespell-project/codespell

`codespell` is one of the leading spell checkers on GitHub.

Going forwards we will need to fix a lot of the misspelled words that are in `codespell.txt`

PR #238, opened 2024-12-31
Remove unneeded blank lines from cxx files

jbampton:


main/automation/source/server/recorder.cxx
```diff
@@ -586,7 +586,7 @@ IMPL_LINK( MacroRecorder, EventListener, VclSimpleEvent*, pEvent )
 					if ( pTB->GetUniqueOrHelpId() != nUId )	// Also Button auf der ToolBox gefunden
 					{
 						if ( nParams == PARAM_NONE )
-						{			// Wir f�lschen einen Parameter
+						{			// Wir f�lschen einen Parameter
```
jbampton:
```suggestion
						{			// Wir fälschen einen Parameter
```

main/automation/source/server/recorder.cxx
Pilot-Pirx, line 1: Yes, still old German comments from the time where no Unicode was used... ;-)

jbampton: Some of these German words have found their way into our codespell ignored words list in #191

Pilot-Pirx: Often German Umlauts were replaced:
ä -> ae
ö -> oe
ü -> ue
ß -> ss
I also saw:
ä -> "a
ö -> "o
ü -> "u
Basically, we should translate these comments finally... ;-)

Pilot-Pirx (approved):

PR #242, opened 2025-01-01
Fix end of hxx files remove unneeded blank lines

jbampton:


jbampton:
This was an auto fix using pre-commit but I just submitted the line ending changes previously.

From the official pre-commit hooks there is the `end-of-file-fixer` which was used:

https://github.com/pre-commit/pre-commit-hooks?tab=readme-ov-file#end-of-file-fixer

I can see the diff is too large / too many files etc for proper review.

We could start with a subset of files ?

Pilot-Pirx:
It was the massive amount of files why I didn't merge it.
But in the end, we can always revert the commit...

PR #248, opened 2025-01-10
Make shell scripts executable with `chmod +x`

jbampton:
`find . -name "*.sh" -execdir chmod u+x {} +`

https://stackoverflow.com/questions/44798121/how-to-find-all-sh-files-and-make-them-executable-using-bash-in-linux


Pilot-Pirx:
Hi Arrigo,
I remember that we had issues (permission) with some scripts.
Can you have a look?

PR #255, opened 2025-01-24
Add more official pre-commit hooks

jbampton:
https://github.com/pre-commit/pre-commit-hooks


Pilot-Pirx:
I would also be interested in:
https://github.com/pre-commit/pre-commit-hooks?tab=readme-ov-file#mixed-line-ending

I already found and fixed some files with mixed line endings, but I could imagine that there are more of them...

jbampton:
I pushed a second commit that fixed the line mixed line ending in `main` for just `cxx` and `hxx` files.

Was a large amount just for this batch

PR #256, opened 2025-01-24
Add basic EditorConfig file

jbampton:
https://editorconfig.org/

From the site:

"EditorConfig helps maintain consistent coding styles for multiple developers working on the same project across various editors and IDEs.
 The EditorConfig project consists of a file format for defining coding styles and a collection of text editor plugins that enable editors to read the file format and adhere to defined styles.
 EditorConfig files are easily readable and they work nicely with version control systems."

Most editors / IDEs are supported

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,13 @@
+# EditorConfig is awesome: https://editorconfig.org
+
+# top-most EditorConfig file
+root = true
+
+# Unix-style newlines with a newline ending every file
+[*]
+end_of_line = lf
+insert_final_newline = true
+
+# Tab indentation (no size specified)
+[Makefile]
+indent_style = tab
--- a/.editorconfig
+++ b/.editorconfig
@@ -7,6 +7,7 @@ root = true
 [*]
 end_of_line = lf
 insert_final_newline = true
+tab_width = 4
 
 # Tab indentation (no size specified)
 [Makefile]


jbampton:
There is a good list of projects using EditorConfig:

https://github.com/editorconfig/editorconfig/wiki/Projects-Using-EditorConfig

ardovm:
I suggest we add:
`tab_width = 4`
to the `[*]` section because many files are formatted this way.

[A page on our wiki](https://wiki.openoffice.org/wiki/Cpp_Coding_Standards/FORMAT) suggests to use 4 spaces for indentation, so we could also add:
```
indent_style = space
indent_size = 4
```
to the same section.

Pilot-Pirx:
Wiki is often outdated.

Would that correspond to:

/* vim: set noet sw=4 ts=4: */

ardovm:
> Wiki is often outdated.

...but I liked it that way... ;-)

In fact, I thought that that wiki page was written _after_ many files were already formatted with tabs.

It may be worth a short discussion on dev@.

> Would that correspond to:
> 
> `/* vim: set noet sw=4 ts=4: */`

This should represent the current state of many files, i.e. tabs, to be displayed as 4 spaces.

I was suggesting we rather set `et` instead of `noet`, using no tabs for indentation, but rather 4 spaces.

ardovm:
> Wiki is often outdated.
> 
> Would that correspond to:
> 
> /* vim: set noet sw=4 ts=4: */

I looked further and now I agree that this is our current standard. Yes, the wiki is outdated.

So this PR shall only add:
```
tab_width = 4
```

.editorconfig
```diff
@@ -0,0 +1,13 @@
+# EditorConfig is awesome: https://editorconfig.org
+
+# top-most EditorConfig file
+root = true
+
+# Unix-style newlines with a newline ending every file
+[*]
+end_of_line = lf
+insert_final_newline = true
+
```
jbampton, line 11:
```suggestion
tab_width = 4

```

Pilot-Pirx:
Ready to be merged? As I understand it this only changes the defaults for the editor on GitHub?

jbampton:
> Ready to be merged? As I understand it this only changes the defaults for the editor on GitHub?

Hey @Pilot-Pirx no this applies to your local editor or IDE and most recognize the format with no plugin needed.

For example I think you use Vim so Vim is supported automatically:

https://editorconfig.org/#pre-installed

Also thinking now we need to add `*.mk` as tabs

PR #262, opened 2025-01-28
pre-commit end-of-file-fixer check more file types

jbampton:
Seems just the java files were cleaned up


jbampton:
I can make a smaller PR with less files if needed ?

PR #277, opened 2025-02-11
pre-commit: add hook to auto trim trailing whitespace

jbampton:
So far targets only `.pl` files

https://github.com/pre-commit/pre-commit-hooks?tab=readme-ov-file#trailing-whitespace


Pilot-Pirx:
That looks good!
It removes all the spaces and tabs at the end of a line. Our code is really badly formatted... ;-)

PR #294, opened 2025-02-16
Update statemnt.cxx - [FIX] successfull

BaseMax:
```
if ( bItemFound )	// FIND_ITEM successfull
```

to

```
if ( bItemFound )	// FIND_ITEM successful
```


Pilot-Pirx:
Looks good to me!
In the end all those German comments must be translated but that should be done by a native speaking developer...

Pilot-Pirx:
Thanks again for your PR.
Are you able to locate/fix these bad encoded German Umlaute in other files?

BaseMax:
Many thanks.

> Are you able to locate/fix these bad encoded German Umlaute in other files?

Of course, I will work on it. `Umlaute`

PR #322, opened 2025-03-02
pre-commit: add official hook fix-byte-order-marker

jbampton:
https://github.com/pre-commit/pre-commit-hooks#detect-aws-credentials

https://github.com/pre-commit/pre-commit-hooks#fix-byte-order-marker

Both checks/hooks/tests already pass.  Good to have some extra tests going forwards.


Pilot-Pirx:
Regarding BOM:
I think I removed all of them (hopefully)

jbampton:
Yes the `fix-byte-order-marker` hook passed.

I think it is still good to have this check so that we don't regress backwards.

The run still takes about two minutes.

Pilot-Pirx:
Definitely!
I did all that "by hand" ;-)

dave2wave:
I don't think we have any use of aws cli. How is the detect-aws-credentials check helpful for us? Is there a better check that is similar?

jbampton:
Hi @dave2wave  I think the AWS check/hook would fail if it finds AWS credentials in the repo.  

So then for us it would be like a security check locally to warn when AWS credentials are found.   Maybe some misplaced credentials etc.

A minor check / test.  I can remove this if needed.

https://github.com/pre-commit/pre-commit-hooks?tab=readme-ov-file#detect-private-key

There is also another hook called detect-private-key

dave2wave:
https://github.com/pre-commit/pre-commit-hooks/blob/main/pre_commit_hooks/detect_aws_credentials.py

It looks like this is more about the environment that the precook is running in? I think this does nothing for us unless it is for CI.

https://github.com/pre-commit/pre-commit-hooks/blob/main/pre_commit_hooks/detect_private_key.py

The detect-private-key test looks interesting for another PR.

.pre-commit-config.yaml
```diff
@@ -32,8 +32,11 @@ repos:
       - id: check-illegal-windows-names
       - id: check-merge-conflict
       - id: check-vcs-permalinks
+      - id: detect-aws-credentials
+        args: [--allow-missing-credentials]
```
jbampton:
```suggestion
```

PR #323, opened 2025-03-02
Create .gitattributes

BaseMax:


jbampton (commented):
@Pilot-Pirx  for reference:

https://docs.github.com/en/get-started/git-basics/configuring-git-to-handle-line-endings

PR #327, opened 2025-03-04
pre-commit(mixed-line-ending): autofix `.c` files

jbampton:


Pilot-Pirx:
I am not sure how Yuri builds OpenOffice on OS/2, but I think he uses a bash shell.
OS/2 uses (like Windows) CRLF as line endings, but the available editors should be able to handle both.

PR #328, opened 2025-03-07
Fix for issue 121222

oooforum:
Treat report chart in ORB extension as a draw chart


Pilot-Pirx:
Can anyone verify this fix?

PR #330, opened 2025-03-10
Fix spelling

jbampton:


Pilot-Pirx:
I am still not sure if "somewhere wishes" should be "someone wishes"?
But we can discuss such things later... ;-)

PR #342, opened 2025-03-30
pre-commit clean ups for `.el` and `.pas` files

jbampton:
Enforced 3 hooks:

- end-of-file-fixer
- mixed-line-ending
- trailing-whitespace

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -33,12 +33,12 @@ repos:
       - id: check-merge-conflict
       - id: check-vcs-permalinks
       - id: end-of-file-fixer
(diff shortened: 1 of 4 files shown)


Pilot-Pirx:
Regarding the .pas files:
Is it OK to add the ASF license when it is already licensed under BSD? Or should the BSD license be removed?
I think the ASF licenses were added in an automated way at that time, maybe this was an error.

dave2wave:
> I think the ASF licenses were added in an automated way at that time, maybe this was an error.

Andrew Rist was the individual from Oracle who made the license header changes to the ALv2.0. I'll assume that what he did was correct and that the intention was to dual license these two files. This is fine.

dave2wave (approved):

Pilot-Pirx:
Thanks!

PR #361, opened 2025-04-25
misc: fix spelling

jbampton:
General typo fixes and updated the ignored words list


Pilot-Pirx:
Gitbox is not synchronizing again, so the cherry-pick to AOO42X must wait until it works again...

PR #366, opened 2025-05-23
Backporting UI changes for enabling AES-256 encryption to AOO41X

Pilot-Pirx:
Working towards backporting UI changes for enabling AES-256 encryption to AOO41X.


Pilot-Pirx:
One problem I see here is that the changes include new strings. So for en-US it should work, but all other translated versions would have English text (at best).
I will do a test build and see how the German version looks like.

Pilot-Pirx:
Yes, the new strings are there, but as expected in English.
Translation would need patching the SDF files.
IMHO we should not wait longer to release this option, it is almost 11 years old!
https://bz.apache.org/ooo/show_bug.cgi?id=125083

dave2wave (approved):
This feature is important and we should provide this to our users. Let's not let translation efforts block as there is work to do there as well.

Pilot-Pirx:
I am doing a build with patched German translation. If this works well I will commit at least the German SDF. We can always work on other languages afterwards.

Pilot-Pirx:
Test build (en-US):
https://www.dropbox.com/scl/fi/esut6i2gzbnc5id0cy6qe/Apache_OpenOffice_4.1.16_Win_x86_install_en-US_aes.exe?rlkey=qwldaemlt9nszuj0fsunyu8cx&st=p1gck28l&dl=0

Pilot-Pirx:
Test build en-US:
https://www.dropbox.com/scl/fi/esut6i2gzbnc5id0cy6qe/Apache_OpenOffice_4.1.16_Win_x86_install_en-US_aes.exe?rlkey=qwldaemlt9nszuj0fsunyu8cx&st=hlzjtoe7&dl=0

Pilot-Pirx:
Doing a fresh installation the AES-256 encryption is *not* set as default. Users can decide if they want to activate it.
We need to announce the option in our Release Notes when releasing AOO 4.1.16.

pedlino (approved):
Looks good to me!

pedlino (commented):
Perfect translation!

Pilot-Pirx:
Hehe, I took the strings from AOO42X, so maybe you did the job? ;-)

PR #368, opened 2025-05-28
Fix spelling

jbampton:
General spelling clean up in many different file types

Updated the codespell ignored words list


Pilot-Pirx:
I had to revert the fix in "main/graphite/graphite-2.3.1.patch" because it could not be applied while building. The typo is in the original graphite code. ;-)

PR #375, opened 2025-06-30
pre-commit: auto remove trailing whitespace from `.py` files

jbampton:
Ran pre-commit locally to clean up the whitespace


gmcdonald (commented):
its fine to remove the whitespace frm the .py files, but I think the py file type should remain in the precommit yaml as it is designed to prevent future occurances.

PR #376, opened 2025-06-30
misc: fix typos

jbampton:
Update ignored words list


gmcdonald (commented):
Isn't the purpose of the codespell linter file to have these mis-spellings on purpose?

jbampton:
> Isn't the purpose of the codespell linter file to have these mis-spellings on purpose?

The file `codespell.txt` contains the misspelled words or terms in the codebase.

So when I fix the spelling in the codebase I then recreate the `codespell.txt` file and this file becomes smaller with less misspelled words.

So you can see an example of how the "ignored words list is created" here:

https://github.com/apache/myfaces/pull/917

PR #386, opened 2025-07-18
pre-commit auto remove trailing whitespace from C files

jbampton:
Another big whitespace clean up


Pilot-Pirx:
Yes, most of them are the License Headers which were added automatically at some time.
And all the others are "historical" since the code in parts is 30 years old. ;-)

PR #390, opened 2025-07-24
pre-commit auto whitespace cleanup for cfg files

jbampton:
Enforce 3 hooks for cfg files


main/framework/qa/unoapi/testdocuments/delete.cfg
Pilot-Pirx, line 1: I am not sure here:
It looks like a ZIP file?

jbampton: I must have missed that.  Seems to have a "zip icon" when I download the file

main/qadevOOo/testdocs/delete.cfg
Pilot-Pirx, line 1: Same here...

Pilot-Pirx (changes requested):
I think we should remove the cfg files that are ZIPs (from this PR).

Pilot-Pirx:
The question is, why are the two .cfg files here technically a ZIP?

PR #393, opened 2025-08-04
Add pre-commit hook oxipng for lossless compression of PNGs

jbampton:
This hook is very CPU intensive so we don't want to add it to the CI.

So I set the hook stage to manual and you can do a full repository clean up locally from the repo root by running:

`pre-commit run --all-files --hook-stage manual`

So I ran the oxipng clean up and we now have less data, smaller file sizes for the same quality

https://github.com/oxipng/oxipng

https://github.com/oxipng/oxipng?tab=readme-ov-file#git-integration-via-pre-commit

https://en.wikipedia.org/wiki/Lossless_compression


Pilot-Pirx:
Reducing the size of AOO is always a good idea!
Can you estimate how much smaller it would get?

jbampton:
I used the following commands from "Google" and if correct it makes it about 2 megabytes smaller 

```
openoffice$ git branch
  add-basic-editorconfig
  add-license-hook-yaml
  add-script-for-ignored-words
  chmod+x-shell-scripts
  pre-commit-add-oxipng
* trunk
  whitespace-cleanup-cfg
  whitespace-fixups-2
  whitespace-fixups-pmk
openoffice$ find . -type f -iname "*.png" -print0 | du -ch --files0-from=- | tail -n 1
54M     total
openoffice$ git checkout pre-commit-add-oxipng 
Switched to branch 'pre-commit-add-oxipng'
Your branch is up to date with 'origin/pre-commit-add-oxipng'.
openoffice$ find . -type f -iname "*.png" -print0 | du -ch --files0-from=- | tail -n 1
52M     total
openoffice$ 
```

PR #394, opened 2025-08-04
Add pre-commit hook to auto insert license headers for Markdown

jbampton:
Ran the hook the first time it auto inserted the license headers and failed.

Second time ran the hook it passed and no files were changed.

The Lucas-C hooks are used on both Airflow and Sedona.


jbampton:
All checks pass. Only Markdown files modified adding license headers.

Good to merge. Thanks

dave2wave:
It is not required to add ASF License files to Markdown. In fact it may get in the way of rendering. What is acceptable is to include metadata that points to the license.

Please DO NOT apply a similar PR to either openoffice-org or openoffice-project

PR #401, opened 2025-08-22
Remove leftover from Mozilla address book

Pilot-Pirx:


Pilot-Pirx:
We should at least add code for Thunderbird address book.

PR #422, opened 2025-09-09
python: remove tabs and indent with spaces

jbampton:
https://peps.python.org/pep-0008/#indentation

PEP 8 – Style Guide for Python Code

Use 4 spaces per indentation level.

https://peps.python.org/pep-0008/#tabs-or-spaces

Python disallows mixing tabs and spaces for indentation.


Pilot-Pirx:
Maybe we should add some kind of vim modeline to Python files?

jbampton:
We could use black to standardize all the Python code to one format.

Then we don't have worry about nit picking in future and just focus on real changes.

Might be similar to adding pre-commit and take a while over many PRs targeting different folders or code etc.

Probably not advisable to just mass format in one go.

We can add black to pre-commit and I have done this already on other projects

https://black.readthedocs.io/en/stable/

https://github.com/psf/black

PR #427, opened 2025-09-29
Fix update-related issues due to GUI functions in secondary threads

ardovm:
The following bugs:
- [127154](https://bz.apache.org/ooo/show_bug.cgi?id=127154)
- [128617](https://bz.apache.org/ooo/show_bug.cgi?id=128617)

and possibly others are caused by GUI calls in secondary threads.

The purpose of this pull request is to fix the above and similar bugs.


ardovm:
This should fix bug [128617](https://bz.apache.org/ooo/show_bug.cgi?id=128617).

PR #448, opened 2025-12-01
Remove Help Cursor

Pilot-Pirx:
With trunk/AOO42X we should consider, if we still want to support very old Windows versions.
Even ReactOS is based on Windows 2000 (2003?).


main/vcl/inc/win/salids.hrc
```diff
@@ -23,92 +23,89 @@
 #define _SV_SALIDS_HRC
 
 // Cursor
-#define SAL_RESID_POINTER_NULL						10000
-#ifndef W40ONLY
-#define SAL_RESID_POINTER_HELP						10001
```
ardovm, line 28: I think a comment would be useful here, telling that ID 10001 was once used for `SAL_RESID_POINTER_HELP`, that was deprecated in December 2025.

ardovm (changes requested):
If you make sure it compiles, then I only have a minor edit to suggest.

main/vcl/inc/win/salids.hrc
Pilot-Pirx: Good point!

It builds for me...
I changed the Help Cursor some month ago and there was no change on Windows 10 or ReactOS. Let's keep it in AOO41X.

ardovm (approved):

ardovm:
IMHO this PR is ready for merging. I'd squash all commits into one.

PR #451, opened 2025-12-10
New design for table selection cursors

Pilot-Pirx:
At least for Windows we could use 32-bit color-depth cursors (incl. transparency), which are supported since Windows XP.
That should be possible for macOS (PNG), too?
I am sure that we could use PNG files for creation of the X11 cursors, too? But that may need some code changes...


Pilot-Pirx:
@ardovm Could you try to build with the X11 cursor?
It is done completely "by hand", the process is very error prone... ;-)

BTW: Do you have any idea how we could use more "modern" cursors for Linux?

ardovm:
cursor_down
This is how the down pointing arrow appears... I am afraid this is not the outcome you expected?

ardovm:
cursor_right
This is the right-pointing cursor

Pilot-Pirx:
Those are the ones with 24x24. Maybe AOO does not like it and I need to do them 32x32?
How does the south-east look like (16x16)?

Nope, I switched the bits... ;-)

Pilot-Pirx:
These X11 cursors are basically XBM files (monochrome cursor). Maybe we could switch to XPM at some point?

ardovm:
On Linux, your cursors appear as black triangles. The PNG's you uploaded for aqua seem to be white triangles instead... is this intended?

Pilot-Pirx:
Yes, Linux (and OS/2) cursors are black.
However, the "mask" appears as white.

ardovm (approved):
Looks good to me on Linux.
I suggest squashing the commits into a single one.

PR #457, opened 2026-01-09
Update cursors for macOS

Pilot-Pirx:
Update some cursors, remove some icons.


Pilot-Pirx:
@jimjag
As we already use PNG with transparency as cursors for macOS we could use grayscale instead of black/white to make them look nicer. This is just a Proof of Concept to see if it builds.
(Additionally I removed 3 icons that I think are not used at all)
If it works I will try to recreate more icons and use them for Windows and macOS.

Pilot-Pirx:
Personally, I think it is no risk to merge this PR.

But it would still be nice if someone could try a build for macOS.

PR #470, opened 2026-02-04
Build Win32 setup only with Unicode support.

ardovm:
Non-unicode ("ANSI") functions were apparently needed only for supporting Windows 9X, that is not our priority any more.

This simplifies the internal structure of the setup executable.


Pilot-Pirx:
I fixed some empty lines at the file ending...

Pilot-Pirx:
I tried to build but it complains about a missing "setup_main.hxx"?

ardovm:
> I tried to build but it complains about a missing "setup_main.hxx"?

This must be because file `setup_main.hxx` was deleted by the commit, but temporary build files on your system still refer to it.

Please try removing directory `main/desktop/wntmsci12.pro`.

Pilot-Pirx:
Builds and installs now.

Pilot-Pirx:
Could this probably fix this bug?
https://bz.apache.org/ooo/show_bug.cgi?id=128593

ardovm:
> Could this probably fix this bug? https://bz.apache.org/ooo/show_bug.cgi?id=128593

Unfortunately, I don't think so... this PR is about removing code that was supposed to run only under Windows 9x.

PR #483, opened 2026-06-07
Fix issues with SELinux

ardovm:
Newer Linux distributions come with SELinux on by default.
Binaries produced by our "official" build VM do not take that into account.

This pull request is against the AOO41X branch because AOO42X and trunk are supposed to be built with more recent compilers, that do not need this additional invocation of `execstack`.

### DIFF
--- a/main/redland/raptor/makefile.mk
+++ b/main/redland/raptor/makefile.mk
@@ -101,6 +101,10 @@ LDFLAGS:=-Wl,-rpath,'$$$$ORIGIN:$$$$ORIGIN/../ure-link/lib' -Wl,-noinhibit-exec
 LDFLAGS:=-Wl,-R'$$$$ORIGIN:$$$$ORIGIN/../ure-link/lib'
 .ENDIF                  # "$(OS)$(COM)"=="SOLARISC52"
 
+.IF "$(OS)"=="LINUX"
+LDFLAGS+:=-Wl,-z,noexecstack
+.ENDIF # "$(OS)"=="LINUX"
+
 .IF "$(COM)"=="C52" && "$(CPU)"=="U"
 CFLAGS=-m64
 .EXPORT: CFLAGS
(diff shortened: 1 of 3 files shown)


DamjanJovanovic:
Why do we need executable stacks? That seems like a bad idea. Is it the main/bridges code that generates executable snippets of machine code?

ardovm:
I am not sure about what module needs that particular feature.

When running AOO 4.1.16 on systems with SELinux enabled, error messages appear and the audit logs contain entries like the following:
```
time->Sat Jan 10 18:23:08 2026 type=AVC msg=audit(1768065788.356:276): avc:  denied  { execstack } for  pid=5517 comm="soffice.bin" scontext=unconfined_u:unconfined_r:unconfined_t:s0-s0:c0.c1023 tcontext=unconfined_u:unconfined_r:unconfined_t:s0-s0:c0.c1023 tclass=process permissive=0
```

Please see [this thread on dev@](https://lists.apache.org/thread/2y50cgsjcy7l9shm06yt7p8w2s021d6m).

ardovm:
I understood that I was trying to solve the wrong problem.
Stacks shall *not* be  executable!
The culprit seems to be the `raptor` library. This PR now ensures that it is built with `noexecstack` under Linux.

This PR is supposed to be squashed and merged.

main/redland/raptor/makefile.mk
```diff
@@ -101,6 +101,10 @@ LDFLAGS:=-Wl,-rpath,'$$$$ORIGIN:$$$$ORIGIN/../ure-link/lib' -Wl,-noinhibit-exec
 LDFLAGS:=-Wl,-R'$$$$ORIGIN:$$$$ORIGIN/../ure-link/lib'
 .ENDIF                  # "$(OS)$(COM)"=="SOLARISC52"
 
+.IF "$(OS)"=="LINUX"
```
DamjanJovanovic: Should we also test for OS="FREEBSD", or use GUI="UNX" instead?

DamjanJovanovic (changes requested):

main/redland/raptor/makefile.mk
ardovm: IIRC FreeBSD has GCC and Clang... does Clang accept `-Wl,-z,noexecstack`? Otherwise we need to differentiate

DamjanJovanovic: Yes, Clang does accept it.

DamjanJovanovic (approved):

PR #486, opened 2026-06-21
Fix rtl string qa NULL-deref crashes (test bug + source hardening)

leginee:
This PR is the first out of 3 fixes from test migration. The original test did not make sense to me. I added the test for NULL pointers in the code itself as guards. And turned the tests into boundry tests.
The code is ai generated. best is to have more eyes to check it.

# commit message for conveniance
Error case 1 of the test migration: ~40 *_000 cases in the rtl string qa suites passed NULL into C string functions that dereference it (e.g. rtl_str_compare(NULL, NULL)), causing 0xC0000005 AVs. These were dormant under the old dmake build and only surface now that the tests actually run. NULL violates the functions' documented contract ("must be null-terminated"), so the defect was in the tests, not the (correct, unchanged) source. Fixed both sides for defense in depth.

Tests (qa/rtl/ostring/rtl_str.cxx, rtl_string.cxx, qa/rtl/oustring/rtl_ustr.cxx):
- Rewrote the UB NULL-deref cases as contract-respecting boundary tests (empty string, prefix/ordering-sign < 0 / > 0), which also closes a previously-untested coverage gap (result sign was never asserted).
- Added real assertions to the safe NULL-with-length-0 cases, which document the length-bounded functions' tolerance as a regression guard.

Source (rtl/source/strtmpl.c, ustring.c):
- Added entry-point NULL guards: OSL_PRECOND (diagnoses misuse loudly in non-product builds, compiles out in product builds) plus a defined empty-string fallback so the library never dereferences NULL. Guards sit at function entry, outside the per-character loops, so string-processing throughput is unchanged.
- strtmpl.c: one edit covers both the sal_Char and sal_Unicode instantiations. getLength is the choke point (guarding it transitively protects hashCode, lastIndexOf*, indexOfStr, trim); compare/compareIgnoreAsciiCase/indexOfChar/replaceChar/ toAscii{Lower,Upper}Case/valueOfChar guarded directly.
- ustring.c: guarded the 6 mixed UTF-16/ASCII comparison helpers; length-bounded args clamp the length to 0 to avoid NULL+0 pointer arithmetic.

BUILD.bazel: rtl_str/rtl_ustr/rtl_string removed from the "known upstream failures" notes; they now pass.

Verified: sal3.dll rebuilds (both template instantiations) and rtl_str / rtl_ustr / rtl_string all pass.

### DIFF
--- a/main/sal/qa/rtl/ostring/rtl_str.cxx
+++ b/main/sal/qa/rtl/ostring/rtl_str.cxx
@@ -37,9 +37,10 @@ namespace rtl_str
 
     TEST_F(compare, compare_000)
     {
-        // The former test passed (NULL, NULL).  rtl_str_compare documents that
-        // both strings must be null-terminated, so NULL is undefined behaviour
-        // (the implementation unconditionally dereferences both pointers).
+        // A former version of this test passed (NULL, NULL).  rtl_str_compare
+        // documents that both strings must be null-terminated, so NULL is
+        // undefined behaviour (the implementation unconditionally dereferences
+        // both pointers).
         // Instead verify the ordering-sign contract, which had no coverage:
         // a value < 0 / > 0 depending on which string is "less".
         rtl::OString aStr1 = "abc";
@@ -53,7 +54,8 @@ namespace rtl_str
 
     TEST_F(compare, compare_000_1)
     {
-        // The former test passed (validStr, NULL) which is undefined behaviour.
+        // A former version of this test passed (validStr, NULL) which is
+        // undefined behaviour.
         // Verify the empty-string boundary instead (the valid analogue of an
         // "absent" string): a non-empty string is greater than the empty one.
         rtl::OString aStr1 = "Line must be equal.";
--- a/main/sal/qa/rtl/ostring/rtl_str.cxx
+++ b/main/sal/qa/rtl/ostring/rtl_str.cxx
@@ -37,13 +37,32 @@ namespace rtl_str
 
     TEST_F(compare, compare_000)
     {
-        rtl_str_compare( NULL, NULL);
+        // The former test passed (NULL, NULL).  rtl_str_compare documents that
+        // both strings must be null-terminated, so NULL is undefined behaviour
+        // (the implementation unconditionally dereferences both pointers).
+        // Instead verify the ordering-sign contract, which had no coverage:
+        // a value < 0 / > 0 depending on which string is "less".
+        rtl::OString aStr1 = "abc";
+        rtl::OString aStr2 = "abd";
+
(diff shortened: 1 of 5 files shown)


main/sal/rtl/source/strtmpl.c
```diff
@@ -55,13 +55,63 @@ inline void rtl_str_ImplCopy( IMPL_RTL_STRCODE* pDest,
     }                                                               \
 }
 
+/* ======================================================================= */
+/* NULL-pointer guards                                                     */
+/*                                                                         */
```
DamjanJovanovic, line 60: Please remove these NULL-pointer guards in release mode for better performance, and only use them in debug builds.

DamjanJovanovic (changes requested):

main/sal/rtl/source/strtmpl.c
leginee: do you think their performance impact is that visible?

leginee: @DamjanJovanovic : switches added.

main/sal/qa/rtl/ostring/rtl_str.cxx
```diff
@@ -37,13 +37,32 @@ namespace rtl_str
 
     TEST_F(compare, compare_000)
     {
-        rtl_str_compare( NULL, NULL);
+        // The former test passed (NULL, NULL).  rtl_str_compare documents that
```
ardovm: "The former test" is IMHO unclear.
We could better phrase it as "A former version of this test"

main/sal/qa/rtl/ostring/rtl_str.cxx
```diff
@@ -37,13 +37,32 @@ namespace rtl_str
...
+        ASSERT_TRUE(rtl_str_compare( aStr1.getStr(), aStr2.getStr()) < 0)
+            << "\"abc\" must compare less than \"abd\".";
+        ASSERT_TRUE(rtl_str_compare( aStr2.getStr(), aStr1.getStr()) > 0)
+            << "\"abd\" must compare greater than \"abc\".";
     }
 
     TEST_F(compare, compare_000_1)
     {
+        // The former test passed (validStr, NULL) which is undefined behaviour.
+        // Verify the empty-string boundary instead (the valid analogue of an
```
ardovm, line 59: Ditto: "A former version of this test..."

leginee:
@DamjanJovanovic I added an else part, which will remove the guards in prod.
Is this satisfactory or should i find a different way?

DamjanJovanovic (approved):

PR #488, opened 2026-06-21
Fix sal encoding-dispatch

leginee:
This PR is the second out of 3 fixes from test migration. This is about the ISCII encoding has been removed. The removal resulted in a array to enumn mismatch. After the enumn and the array fitted together, some encoding typos appeared.
enumn has been untouched for ABI compatibility on our Interface.

# commit message for convenience
Fix sal encoding-dispatch off-by-one (ISCII removal) + ZapfDingbats table bugs

Impl_getTextEncodingData() is a positional array indexed by the ABI-frozen rtl_TextEncoding enum. #i119141 removed the ISCII converter but deleted its array slot with no NULL placeholder, shifting JAVA_UTF8, ADOBE_STANDARD, ADOBE_SYMBOL, PT154 and ADOBE_DINGBATS onto the wrong converter. Restore the NULL slot and add a compile-time size guard.

Wiring ADOBE_DINGBATS correctly surfaced three latent ZapfDingbats table errors (0x6C→U+25CF; reverse range to U+2797; circled-digits reverse start 0xAC); fix all three. Update rtl_textcvt qa: park the now-unimplemented ISCII behind TEST_ISCII_DEVANAGARI, correct the ISCII Windows codepage to 57002.


leginee:
would be great if someone could review. Sorry forgot to add reviewers. :)

DamjanJovanovic (approved):
Would be nice if we could get that converter working again, but alright, remove it for now.

leginee:
I did not remove the converter. This patch fixes an error that happend during the removal and fixes another issue that was shadowed by the issue. :)

PR #490, opened 2026-07-05
Fix building Python3 on linux

ardovm:
These commits should allow to compile Python 3.11.15 on Linux.

However, after applying them, Python macros do not seem to work. But at least, we can get to compile trunk again.

### DIFF
--- a/main/configure.ac
+++ b/main/configure.ac
@@ -4140,7 +4140,7 @@ elif test -n "$with_system_python" -o -n "$with_system_libs" && \
    python_version=`$PYTHON -c "import sysconfig; print(sysconfig.get_config_var('VERSION'));"`
    # Python 3.8 and later distinguish building Python modules from linking to the libpython:
    # (https://bugs.python.org/issue36721)
-   # We cannot use test(1) because Python version numbers are floats
+   # We cannot use test(1) because Python version numbers are in the "major.minor.micro" form
    if $PYTHON -c "import sys; v = sys.version_info; g = ((v.major < 3) or ((v.major == 3) and (v.minor < 8))); sys.exit(not g)"; then
        # < 3.8
        PKG_CHECK_MODULES(PYTHON, python-$python_version, PYTHON_PC="TRUE", PYTHON_PC="" )
--- a/main/configure.ac
+++ b/main/configure.ac
@@ -4140,9 +4140,12 @@ elif test -n "$with_system_python" -o -n "$with_system_libs" && \
    python_version=`$PYTHON -c "import sysconfig; print(sysconfig.get_config_var('VERSION'));"`
    # Python 3.8 and later distinguish building Python modules from linking to the libpython:
    # (https://bugs.python.org/issue36721)
-   if test "$python_version" -lt "3.8"; then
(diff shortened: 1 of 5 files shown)


ardovm:
Requesting @jimjag's review in particular for the name of the `sysconfigdata` file in b0bc83f.

DamjanJovanovic:
Why don't we use the system Python on Linux? Do macros work with it?

ardovm:
I suppose we are not using system-provided Python (and also other additional dependencies) due to the possible age difference between the distro of the build system and the distro of the actual systems on which AOO will run.
I'll see if I can build with system Python.

leginee:
main\pyuno\source\module\pyuno_dlopenwrapper.c 
links against 
#include <Python.h>

Isnt that an issue?

what we need is a dialog similar or same that lists as with java, so we can point python to a sandbox.
That would be cool and not that difficult i think.

ardovm:
> main\pyuno\source\module\pyuno_dlopenwrapper.c links against #include <Python.h>
> 
> Isnt that an issue?

I suppose not. The build system should take care of indicating the proper include file paths: either local to the AOO build tree, or at system level.

leginee:
> I suppose not. The build system should take care of indicating the proper include file paths: either local to the AOO build tree, or at system level.
But we do not build for a specific distro, but in general. And if you build against the distros python, is it compatible to other distros python?

ardovm:
> Why don't we use the system Python on Linux? Do macros work with it?

I just built on openSUSE Leap 16.0, that ships Python 3.13.13. Macros are accessible when using system Python.
Macros are not accessible when using the "bundled" Python 3.11.5.

ardovm:
> > I suppose not. The build system should take care of indicating the proper include file paths: either local to the AOO build tree, or at system level.

> But we do not build for a specific distro, but in general. And if you build against the distros python, is it compatible to other distros python?

I am afraid that using system level Python would make the AOO binary incompatible, or would raise compatibility issues, with other distributions having other system Python versions.

DamjanJovanovic:
> I am afraid that using system level Python would make the AOO binary incompatible, or would raise compatibility issues, with other distributions having other system Python versions.

Maybe. Come to think of it, we only use system Python 3 on FreeBSD too, so what you are doing is really new.

Do hanya's Python 3 patches contain anything helpful? https://bz.apache.org/ooo/show_bug.cgi?id=123975

Oh and please enable debug logging for your internal Python using something like this:
PYSCRIPT_LOG_LEVEL="DEBUG" PYSCRIPT_LOG_STDOUT="1" ./soffice
It might show some helpful output.

jimjag (commented):
LGTM

ardovm:
Exception: "com.sun.star.beans.MethodConcept.ALL is unknown"
...does it ring any bells?

The problem seems to be the initialization of the pyuno module, when it loads file `unohelper.py`

If I _manually_ start the bundled python (using the `program/python` script) it runs `unohelper.py` flawlessly.

jimjag:
This should be fixed in trunk. Closing but re-open if it shows up again.

PR #491, opened 2026-07-07
Bugfix dxf filter

leginee:
Hello all, 
This branch fixes dxf related issue. I set the scope within the filter. There are issues currently open, their root cause are render issues, and not rooted in the filter. To limit impact of the patch I skipped them for now.
Bugzilla is updated accordingly.


ardovm:
Looks good to me!

PR #494, opened 2026-08-12
Update Capitalise.py

jeanmicoste:
Avoid the error described here : [https://forum.openoffice.org/en/forum/viewtopic.php?p=556640#p556640](url)


leginee:
this seems to be from https://bz.apache.org/ooo/show_bug.cgi?id=82861
I will review as soon as i have time, and get this on board. :)

PR #496, opened 2026-08-23
Replace older Boost math with std C++. Update error reporting for CoinMP

pgiffuni:
The solver component was made originally for lp_solve, which did basically the same as CoinMP, but handled differently some errors, in particular timeouts.

Be more in line with CoinMP error reporting. By the way we set up the problem internally it is unlikely some of these will come up but defining them will make easier to understand what could go wrong.


pgiffuni:
This has all been tested on FreeBSD.

PR #497, opened 2026-08-23
Fix non-system build of Python 3 on Linux

ardovm:
This PR adds missing bits of PR #490 and finally enables Python3 macros on Linux, when not using system Python.


ardovm:
Overcome by 91144cd0085

